PR #34, opened 2019-08-28
Add detourInformation field to activity definition common core data

j-d-b:
## Issue
I'm part of a team working with MassDOT to implement the WZDx spec. MassDOT currently uses and would like to disseminate `Detour Information`, though it is not included in the current version of the WZDx spec.

## Proposal
I've updated the common core data to include an optional (for backwards compatibility) `detourInformation` field in both `common-core-dictionary.md` and `full-spec.md`.

## Examples and Current Use
I updated the MassDOT examples to show this field in use. The other examples are still valid as the `detourInformation` field is optional.

### DIFF
--- a/create-feed/massdot.md
+++ b/create-feed/massdot.md
@@ -20,6 +20,7 @@ closedShoulders | inside
 workersPresent | n/a
 RoadRestriction*<ul><li>roadRestriction</li></ul> | n/a
 description | I-91 Southbound   Exit (20) Rt-5/ Rt-10 Northampton Hadley to Exit<br>(19) Rt-9
+detourInformation | Take Exit 21 to I-91SB
 issuingOrganization | MassDOT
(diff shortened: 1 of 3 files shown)


sknick-iastate:
I like this idea but would like to get feedback from data consumers.  Is the intention that this actually provides the detour path or just says a detour is in effect?  I think a yes/no indicator that a detour is in effect is easier to implement but it probably is not much additional effort to provide the path.

I think there may be some benefit to consider the ability to have multiple detours.  Depending on how long the closure is in effect we may have both a global and local detour.  The global is for longer distance travelers to avoid the area while the local is for immediate drivers in the area.

j-d-b:
This needs more discussion before being proposed for implementation. The changes proposed are also not relevant after v2 changes. I will create an issue for discussion.

PR #35, opened 2019-08-28
Add town field to BeginLocation and EndLocation data frames

j-d-b:
## Issue
Currently, there is no notion of the town/city anywhere in the specification. I'm part of a team working with MassDOT, and we believe town (as the general term; also encompasses city) would be beneficial information about a work zone to include in our data feed. Since the work zone activity may span multiple towns, the `town` field should be added to the `BeginLocation` and `EndLocation` data frames individually rather than the common core data.

## Proposal
I've updated the `BeginLocation` and `EndLocation` data frames to include a `town` field. For flexibility and backward compatibility, this field is optional.

The field name *"town"* was chosen over *"city"* as we felt it was a more general term and more familiar to MassDOT. If other DOTs/entities have comments about the naming of this field or if there are other standards and we can come to a different consensus, I will gladly update the PR.

## Examples and Current Use
I updated the MassDOT examples to show the `town` field in use. The other examples are still valid as the `town` field is optional.

### DIFF
--- a/create-feed/massdot.md
+++ b/create-feed/massdot.md
@@ -10,8 +10,8 @@ Tag | Value
 Identifier | 137097
 StartDateTime<ul><li>startDateTime-ver</li></ul> | startDateTime-ver: 2016-11-03T19:37:00
 EndDateTime<ul><li>endDateTime-est</li></ul> | endDateTime-est: 2016-11-04T05:30:00
-BeginLocation<ul><li>roadName</li><li>roadNum</li><li>roadDirection</li><li>latitude</li><li>longitude</li><li>milepost</li><li>town</li></ul> | roadName: I-91<br>roadDirection: southbound<br>latitude-est: 42.33865<br>longitude-est: -72.63399<br>town: Northampton
-EndLocation<ul><li>latitude</li><li>longitude</li><li>milepost</li><li>town</li></ul> | latitude-est: 42.33307<br>longitude-est: -72.6214<br>town: Northampton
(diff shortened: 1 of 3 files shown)


Polly-Okunieff:
Two comments with respec to this issue: 
(1) The work zone may also cross multiple states.  
(2) "Town" may be too limiting. Suggest "jurisdiction" be used instead.

k-vaughn:
I do not know if all states have these unique codes currently, but they will need these codes in the near future to allow proper operation of IEEE 1609.2 security certificates. 

For example, connected vehicles need to know that the map and signal phase and timing messages are coming from an authorized source. These messages must be signed by a local agency using its own code. This identifier is defined in the current draft of ISO 17419-2 as an object identifier, which basically means it is a sequence of integers where each integer represents a node in a hierarchy. Thus, the likely way this plays out is that each jurisdiction will have a code of the format us.texas.harris-county.houston where each Hierarchical jurisdiction is represented with an integer value. 

Note: it is up to each jurisdiction to determine how sub-jurisdictions are assigned and numbered. (e.g., a small country like Singapore might not have as many levels in its hierarchy as the US.)

If the WZDx needs similar codes, it is probably worthwhile reusing the same code rather than inventing a new one.

Regards,
Ken Vaughn

j-d-b:
> (1) The work zone may also cross multiple states.

This is fine as the field is just a label.

> (2) "Town" may be too limiting. Suggest "jurisdiction" be used instead.

Agree; I like this better. I will change the PR following.

j-d-b:
This discussion should occur in an issue prior to a pull request. The changes proposed in this PR are also not relevant to v2. I will create an issue for discussion of the "town" field to continue this discussion.

PR #36, opened 2019-08-28
Add traffic information and free flow values

j-d-b:
## Issue
I'm part of a team working with MassDOT to implement the WZDx spec for our 3rd party work zone data feed. For some work zones, our system has traffic data information from roadside sensors/cameras about delay, travel time, speed, queue length, volume, and more.

We believe it would be beneficial to disseminate a subset of this traffic information to third parties. All the fields I've added in this pull request are optional to ensure backwards compatibility.

## Proposal
Primarily, I've created a new data frame, `trafficInformation`, which contains four traffic data metrics describing current conditions in the work zone activity:
* `averageSpeed`: Average speed in mph
* `travelTime`: Travel time in seconds
* `delay`: Delay in seconds
* `queueLength`: Length of queue in miles

Additionally I've added two fields to the common core data:
* `freeFlowTravelTime`
* `freeFlowSpeed`

These may be helpful even without traffic information, but are critical when there is traffic information, so are included in this PR.

## Examples
I've updated the MassDOT examples (e.g. `create-feed/massdot.md`) to demonstrate these fields in use.

*Note PR #24 which preceded this PR*

### DIFF
--- a/create-feed/massdot.md
+++ b/create-feed/massdot.md
@@ -19,6 +19,8 @@ openLanes | right2
 closedLanes | left1
 closedShoulders | inside
 workersPresent | n/a
+freeFlowTravelTime | 380
+freeFlowSpeed | 65
 RoadRestriction*<ul><li>roadRestriction</li></ul> | n/a
 description | I-91 Southbound   Exit (20) Rt-5/ Rt-10 Northampton Hadley to Exit<br>(19) Rt-9
 issuingOrganization | MassDOT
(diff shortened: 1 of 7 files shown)


j-d-b:
I wanted to note that this is certainly not something that can be completed for the v2 release and requires significant further discussion.

j-d-b:
The idea behind these suggested changes should be proposed through an issue, as more discussion is needed to determine the implementation.

PR #54, opened 2019-11-19
Relational data model - clean version

sknick-iastate:
Based on [@j-d-b](https://github.com/usdot-jpo-ode/jpo-wzdx/pull/50#issuecomment-555106423) comments, this pull request is a clean version of the relational data model that @DeraldDudley has developed without the additional changes from other pull requests.  This pull request only changes the structure of the data from version 1.1 to a relational model and very little of the content. All other changes in Pull Request #50 should be made through separate pull requests.

**_A summary of some of the proposed changes:_**
-[Logical Data Model ](https://github.com/sknick-iastate/jpo-wzdx/blob/RelationalDataModel/data-tables/road_event_erd.png)(Entity Relationship Data Model)
-Renamed [WZDx Header](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/data-tables/wzdx-header-data.md) to [road_event_feed_info](https://github.com/sknick-iastate/jpo-wzdx/blob/RelationalDataModel/data-tables/road_event_feed_info.md)
-Renamed [Common Core](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/data-tables/common-core-dictionary.md) to [road_events](https://github.com/sknick-iastate/jpo-wzdx/blob/RelationalDataModel/data-tables/road_events.md)
-Add geometry_type and geometry to replace the begin and end location data frames. Also moved other elements within those data frames to the road_events table.
-Added data elements from the start and end time data frames to the road_events table
-Removed the estimated and verified tags and replaced those with begin location, end location, start time and end time accuracy fields.
-Update the [examples](https://github.com/sknick-iastate/jpo-wzdx/tree/RelationalDataModel/create-feed) in the create-feed folder (would be beneficial to have some one review this in detail)
-Removed the [full_spec](https://github.com/usdot-jpo-ode/jpo-wzdx/tree/master/full-spec) since is was redundant and could result in discrepancies.  All content within the full_spec that was not located within github was added to the top README.md file.
-Update the [sample files](https://github.com/sknick-iastate/jpo-wzdx/tree/RelationalDataModel/sample-files) including xml, json, and xsd

I also addressed some of the comments related to [@j-d-b](https://github.com/usdot-jpo-ode/jpo-wzdx/pull/50#issuecomment-555282214) which I will summarize in the first comment.

[...]

PR #55, opened 2019-11-25
Design version 2 of the work zone data specification to inform drivers about other road events such as closures and detours.

DeraldDudley:
Version 1.1 of the specification fosters information exchange about a single type of road event; work zones.  This pull request makes version 2 of the specification more versatile by enabling information exchange about multiple types of road events such as work zones, closures, detours, and more.  

Including this design feature does not change the focus of this working group; it fosters the specification’s forward-compatibility with a wide array of road events.


j-d-b:
This has been implemented through #54. @DeraldDudley there are a lot of changes here, if anything specifically did not make it into v2 from here, please open a new issue with those changes.

PR #62, opened 2019-12-31
V2 examples editorial

chuehlien:
Changes made:
- Modified the following field names in examples:
  - `type_of_work` --> `types_of_work`
  - `road_num` --> `road_number`
  - `worker_present` --> `workers_present`
- Made sure that the raw geojson example file and the sample info shown in the markdown tables match up.
- Removed some extra brackets.
- Reran JSON through prettify tool.
- *** It'd be best for all Boolean type data to be represented in the same way (all 1/0 or all true/false). Right now, some fields are represented by 1/0 (e.g. `is_architectural_change`) and some are represented by true/false (e.g. `workers_present`). I've changed all boolean fields in the example files to true/false for the time being, but am okay with changing them to 1/0.


j-d-b:
`true`/`false` is the semantic choice for JSON and I'm for using this throughout 👍

chuehlien:
Thanks! I also noticed that all data fields in the examples are stringified regardless of data type (e.g. true/false booleans are represented by "true"/"false", number 15.4 represented by "15.4"). 

Do we want to change the examples to use non-stringified true/false and number values? Or should we keep all fields stringified since the datetime field would remain a string? I'm okay with either option as long as we are consistent.

j-d-b:
> Do we want to change the examples to use non-stringified true/false and number values? Or should we keep all fields stringified since the datetime field would remain a string? I'm okay with either option as long as we are consistent.

We should change these to be the correct/semantic JSON data types--not strings. I had done this in previous cleanup/PRs but perhaps it was overwritten. The JSON types should reflect the data type for the field.

DeraldDudley:
> > Do we want to change the examples to use non-stringified true/false and number values? Or should we keep all fields stringified since the datetime field would remain a string? I'm okay with either option as long as we are consistent.
> 
> We should change these to be the correct/semantic JSON data types--not strings. I had done this in previous cleanup/PRs but perhaps it was overwritten. The JSON types should reflect the data type for the field.

Agree with @j-d-b; GeoJSON is an extension of JSON so it should be compliant with the JSON spec.

chuehlien:
Thanks @j-d-b & @DeraldDudley ! Will make these changes in the examples.

chuehlien:
@DeraldDudley @j-d-b Are the DateTime fields supposed to be always in local time (e.g. `2018-11-13T20:20:39+01:00`) or in UTC (e.g. `2018-11-13T20:20:39Z`)? Sorry if I missed this in the documentation.

j-d-b:
> Are the DateTime fields supposed to be always in local time (e.g. `2018-11-13T20:20:39+01:00`) or in UTC (e.g. `2018-11-13T20:20:39Z`, which I'd recommend)? Sorry if I missed this in the documentation.

I'm not sure this is specified anywhere--it isn't in the data tables field description at least. Clarifying this in the field descriptions is definitely a good idea. I'm also for using UTC, especially since this is how the data should be stored in the database and the relational model is intended to facilitate that.

sknick-iastate:
> > Are the DateTime fields supposed to be always in local time (e.g. `2018-11-13T20:20:39+01:00`) or in UTC (e.g. `2018-11-13T20:20:39Z`, which I'd recommend)? Sorry if I missed this in the documentation.
> 
> I'm not sure this is specified anywhere--it isn't in the data tables field description (for `start_date` and `end_date` at least. Clarifying this in these descriptions is definitely a good idea. I'm also for using UTC, especially since this is how the data should be stored in the database and the relational model is intended to facilitate that.

In the [road events table](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/v2/data-tables/road_events.md) we say in the notes: _All date/time formats shall use ISO 8601 Data elements and interchange formats – Information interchange_

This doesn't require UTC but that is how we are providing the data currently in Iowa.

chuehlien:
> > > Are the DateTime fields supposed to be always in local time (e.g. `2018-11-13T20:20:39+01:00`) or in UTC (e.g. `2018-11-13T20:20:39Z`, which I'd recommend)? Sorry if I missed this in the documentation.
> > 
> > 
> > I'm not sure this is specified anywhere--it isn't in the data tables field description (for `start_date` and `end_date` at least. Clarifying this in these descriptions is definitely a good idea. I'm also for using UTC, especially since this is how the data should be stored in the database and the relational model is intended to facilitate that.
> 
> In the [road events table](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/v2/data-tables/road_events.md) we say in the notes: _All date/time formats shall use ISO 8601 Data elements and interchange formats – Information interchange_
> 
> This doesn't require UTC but that is how we are providing the data currently in Iowa.

Thanks @j-d-b @sknick-iastate - let's explicitly require UTC for all datetime fields for consistency, unless there are objections. I can add that to the field description.

natedeshmukhtowery:
As long as UTC allows for seamless differentiation for daylight savings adoption across jurisdictions (i.e., Arizona, which doesn't do daylight savings and so is "off" from similar jurisdictions half the year), I think it's the way to go.

PR #71, opened 2020-01-24
Full spec

Mahsa-Ettefagh:
This pull request will add full spec file based on the v2.0 published specification to the master branch.


Mahsa-Ettefagh:
Adding Full Spec document to master branch for v2.0

Issue #73, opened 2020-02-20
Clarify optional field representation in GeoJSON feed ouput - create a JSON Schema


j-d-b:
# Clarify optional field representation in GeoJSON feed ouput - create a JSON Schema

## Background
Currently, some WZDx fields are optional. The intention of an optional field is to allow pieces of information that:

* May not apply to a certain road event (e.g. milepost); or
* Are unable to be provided by a significant amount of data producers (e.g. worker presence) 

to be included when relevant or if available.

However, the spec doesn't clarify how an optional piece of data that is intended to have no value should be represented in the GeoJSON feed output. 

## Representing a value-less optional field in the feed output

There are two (reasonable) ways to accomplish this:

1. The JSON property for an optional piece of data (e.g. `beginning_milepost`) is still required to be present in the GeoJSON feed output but its value should be `null` if it is inapplicable/unable to be filled out.
2. The JSON property for an optional piece of data without a value should not be present in the GeoJSON feed output. 

## Implementing the clarification

The chosen representation from the previous section *could* be clarified with an additional paragraph of text in the `create-feed` README, however, a better approach would be to describe the representation by creating a schema for the GeoJSON feed output. A JSON schema leaves  nothing up to interpretation by the reader and has the additional benefit of allowing feed producers to validate their feed output. I believe this benefit makes it worth the additional time to create over just adding a block of explanatory text.

lynnerandolph:
We would definitely like to see this occur!

natedeshmukhtowery:
Agree with the need for a JSON Schema. My only substantive concern at this point would be that a JSON Schema would need to be consistent with what the WZDx specification has tried to balance so far - allowing for the diversity of existing approaches but trying to bring some degree of harmony.

If folks think that a JSON Schema is a helpful approach to feed validation that doesn't become too prescriptive, then it seems to be a very useful, even critical, tool to have. 

Thoughts?

j-d-b:
@natedeshmukhtowery can you clarify what you mean by the following?

> allowing for the diversity of existing approaches but trying to bring some degree of harmony

The JSON Schema would allow for as much diversity of implementation as possible given the required fields and GeoJSON required structure. 

I'm glad to see there was a JSON Schema for v1.1. I feel as though we should definitely go ahead with this for v2. I'll begin putting that together for review.

j-d-b:
Note also we still need community consensus on how we want to represent value-less optional fields, as described in the first issue comment text.

[...]

Issue #70, opened 2020-01-22
Change Time/Spatial Enumerated Type enumerations to be lowercase, for consistency


j-d-b:
# Change Time/Spatial Enumerated Type enumerations to be lowercase, for consistency

Currently, all WZDx enumerated types are all lowercase, dash delineated, except for Geometry Type (which is fine as it is defined by the GeoJSON standard) and time/spatial verification. I don't believe there is a reason that the time/spatial verification enumerated types should be capitalized, so this is unnecessarily inconsistent with the other enumerated types.

sknick-iastate:
Agreed.  We can also remove the (-est) and (-ver) since these are no longer used.  They were previous used in v1.1 for the time and location data frames.

j-d-b:
There are two options for implementing this in v3:

1. Change the Time/Spatial Enumerated Type values to lowercase (i.e. `estimated` and `verified`) and break backwards compatibility
2. Add two new allowed values to the Time/Spatial Enumerated Types and deprecate the uppercase versions, to be removed in a later version

1. is my preference as it's a simple change for producers and makes the v3 spec cleaner

CraigMoore-Sea:
Agree that option 1 is simple even if it break backward compatibility.

sknick-iastate:
This would be an easy change for Iowa to go with option 1.

lynnerandolph:
I agree for TxDOT

j-d-b:
For MassDOT as well, option 1 would be an easy change.

j-d-b:
Fixed in #88

PR #79, opened 2020-04-10
Add JSON Schema for WZDx v2.0 Feed

j-d-b:
This PR resolves issue #73 by adding a `schemas` directory within the `create-feed` directory and a JSON schema for a WZDx v2.0 feed within that directory. The schema is based on JSON Schema Draft 7, which (while not the most current) is the most widely used and supported by validation tools. 

There was discussion in issue #73 about how to represent non-required fields without value, should they be included and given a value of `null` or should the field not be present.

JSON Schema is more conducive to the latter approach, that is, optional properties that the feed producer intends not to include in the feed should not be included on the JSON objects. Thus this was the approach taken for this PR.

One comment I had was regarding the `restrictions` field on the road event, when there are no restrictions this should be an empty array `[]`, which is different than the `restrictions` property not being included which would indicate the restrictions are unknown. This is how I wrote the schema here, however the text in the `restrictions` field description on the `road event` object notes that this field is for "one or more" restrictions. I thus have also changed that text to read "zero or more".


create-feed/schemas/wzdx_v2.0_feed.json
```diff
@@ -0,0 +1,383 @@
...
+      "properties": {
+        "feed_update_date": {
+          "description": "The UTC date and time when the data feed was last updated",
+          "type": "string",
+          "format": "date-time"
+        },
+        "metadata": {
+          "description": "The URL for an associated metadata file",
+          "type": "string",
+          "format": "date-time"
```
enigma-samurai, line 58: Did you mean to put date-time here?

j-d-b: No, this is a typo. Good catch; I will fix this.

create-feed/schemas/wzdx_v2.0_feed.json
```diff
@@ -0,0 +1,382 @@
...
+          "properties": {
+            "$ref": "#/definitions/road_event"
+          },
+          "geometry": {
+            "oneOf": [
+              {
+                "$ref": "https://geojson.org/schema/LineString.json"
+              },
+              {
+                "$ref": "https://geojson.org/schema/MutliPoint.json"
```
chuehlien, line 35: Typo - should be "https://geojson.org/schema/MultiPoint.json"

j-d-b: Very nice find 👏, I had to re-read three times to even see it.

feed-content/data-tables/road_events.md
```diff
@@ -49,8 +49,8 @@ Field Name | Data Type | Description | Conformance | Notes
...
-**restrictions** | Enumumeration; Text | One or more road restrictions applying to the work zone road segment associated with the work zone delimited by semicolons  | Optional | These are included as flags rather than detailed restrictions. Detailed restrictions are coded to specific lanes in the lane_restrictions table. See [Road Restriction Enumerated Type](/feed-content/enumerated-types/road_restriction.md)
-**description** | Text | Short free text description of road event | Optional | This will be populated with formal phrases in a later WZDx version
+**restrictions** | Enumumeration; Text | Zero or more road restrictions applying to the work zone road segment associated with the work zone delimited by semicolons  | Optional | These are included as flags rather than detailed restrictions. Detailed restrictions are coded to specific lanes in the lane_restrictions table. See [Road Restriction Enumerated Type](/feed-content/enumerated-types/road_restriction.md)
+**description** | Text | Short free text description of work zone | Optional | This will be populated with formal phrases in a later WZDx version
```
chuehlien, line 53: Remember to make these changes to the https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/full-spec.md file as well.

create-feed/schemas/wzdx_v2.0_feed.json
```diff
@@ -0,0 +1,382 @@
...
+        "bike-lane",
+        "none",
+        "unknown",
+        "alternating-flow-lane",
+        "outside",
+        "inside",
+        "both"
+      ]
+    },
+    "lane_restriction_unit": {
```
chuehlien, line 376: Update this to "restriction_units" as referenced in lines 261-266 and other documentation, or update the documentation to say "lane_restriction_unit".

create-feed/schemas/wzdx_v2.0_feed.json
```diff
@@ -0,0 +1,382 @@
...
+    },
+    "spatial_verification": {
+      "title": "Spatial Verification Enumerated Type",
+      "description": "An indication of how a geographical coordinate was defined",
+      "enum": ["Estimated", "Verified"]
+    },
+    "time_verification": {
+      "title": "Time Verification Enumerated Type",
+      "description": "A measure of how accurate the a datetime is",
+      "enum": ["Estimated", "Verified"]
```
chuehlien, line 282: To be consistent with other enumerated types, I'd recommend using lowercased enum values (e.g. "estimated" and "verified" as opposed to "Estimated" and "Verified") for "spatial_verification" and "time_verification" enumerated type.

feed-content/data-tables/road_events.md
j-d-b: We're planning on removing the `full-spec.md` file to avoid having to make changes in multiple places going forward. I will update `full-spec.md` in this case to allow this PR to be merged, but I'd strongly advocate for it to be removed before anything else is merged.

create-feed/schemas/wzdx_v2.0_feed.json
j-d-b: See #70. This change will happen for v3.

create-feed/schemas/wzdx_v2.0_feed.json
j-d-b: This is just the name of the JSON schema type definition, not the field on the feed, which is `restriction_units`.  The type does not (and should not always) have to match the field(s) it is used on. In this case, as the types are global to the schema, `lane_restriction_units` is more clear as `restriction_units` doesn't clarify this type is related to lane restrictions only.

feed-content/data-tables/road_events.md
j-d-b: Moreover, it matches the name of the "Lane Restriction Unit Enumerated Type", like the other enumerated types. See, https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/feed-content/enumerated-types/lane_restriction_unit.md

create-feed/schemas/wzdx_v2.0_feed.json
j-d-b: Moreover, it matches the name of the "Lane Restriction Unit Enumerated Type", like the other enumerated types. See, https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/feed-content/enumerated-types/lane_restriction_unit.md

feed-content/data-tables/road_events.md
j-d-b: Updated full-spec.md

create-feed/schemas/wzdx_v2.0_feed.json
chuehlien: Got it - thanks for the explanation! Agree that `lane_restriction_unit` is more clear. In this case, could line 262 be updated from "#/definitions/restriction_units" to "#/definitions/lane_restriction_unit"? I get an error for the schema on https://www.jsonschemavalidator.net/ unless I make that change. Thanks!

j-d-b: Yes good catch again; thanks for your review. This is a typo on my part. Interesting though as I believe I tried it on that validator. I'll fix it now.

create-feed/schemas/wzdx_v2.0_feed.json
```diff
@@ -0,0 +1,382 @@
...
+        "no-trucks",
+        "travel-peak-hours-only",
+        "hov-3",
+        "hov-2",
+        "no-parking",
+        "reduced-width",
+        "reduced-height",
+        "reduced-length",
+        "reduced-weight",
+        "axel-load-limit",
```
chuehlien, line 307: minor typo - "axel-load-limit" -> "axle-load-limit"
per https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/feed-content/enumerated-types/road_restriction.md

create-feed/schemas/wzdx_v2.0_feed.json
```diff
@@ -0,0 +1,382 @@
...
+          "description": "A flag indicating that there are workers present in the road event",
+          "type": "boolean"
+        },
+        "reduced_speed_limit": {
+          "description": "The reduced speed limit posted within the road event",
+          "type": "integer",
+          "minimum": 0
+        },
+        "restrictions": {
+          "description": "One or more road restrictions applying to the road event",
```
chuehlien, line 150: Update this to say "Zero or more" instead of "One or more" to reflect the changes made to "feed-content/data-tables/road_events.md" in this PR.

Great work with the PR overall!!

create-feed/schemas/wzdx_v2.0_feed.json
j-d-b: Your review has been invaluable, great work.

create-feed/schemas/wzdx_v2.0_feed.json
j-d-b: Thank you. Great work with your review and finding my typos. When wrote this schema I kept thinking, "I'm definitely going to make a typo and not be able to find it".

Resolved this one as well.

j-d-b: Or four typos...

davidcraig4300:
Approve these changes

Issue #80, opened 2020-04-13
Format of "road_event_feed_info" "version" property, possible enumerated type


j-d-b:
The `version` property on the `road_event_feed_info` entity is described as 

> The specification version used to create the dataset

This is a text field with no further indication of how the version should be represented, that is, `v2.0`, `V2`, `2`, `2.0`, etc.

This should be standardized, either with clarifying text + a regex in the JSON schema or by creating an enumeration for all the possible feed versions.

My choice for a format for this field would be `2` if we are going to continue release only major versions, and `2.0` if we want to consider minor releases.

Mahsa-Ettefagh:
Comments received during the [V3 Specification Update Subgroup](https://github.com/usdot-jpo-ode/jpo-wzdx/wiki/WZDx-v3-Specification-Update-Subgroup-Meeting-Notes,-2020-04-29) meeting on 4/29/2020:

- Lynne Randolph: I'm good with major.minor
- Neil Boudreau, MassDOT: I think that makes the most sense
- Derald W Dudley: +1 for major.Minor
- Marysa Myers: Agreed, major.minor would be easiest and could avoid any inconsistencies using V, v, v., etc
- Jim Williams (INRIX): Also agree with major.minor.
- Ariel and David Craig to review the definition of major/minor versions of the specification

j-d-b:
Resolved in #89

PR #89, opened 2020-05-08
Require major.minor format for road_event_feed_info `version` field

j-d-b:
This PR resolves issue #80 by noting that the `version` field on the [road_event_feed_info](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/feed-content/data-tables/road_event_feed_info.md) should follow the `major.minor` (e.g. `1.1`, `2.0` format).

~This will be required by the JSON schema (#79) once it is completed and merged in.~
Completed in e26ffa8


sknick-iastate:
+1

j-d-b:
Note #90 may require modifying this. We could however just not include the patch version here as nothing in the spec/output changes when the patch version changes.

j-d-b:
I'm wondering if we should use this opportunity to also require this field, as it's currently optional.

create-feed/schemas/wzdx_v3.0_feed.json
```diff
@@ -57,8 +57,9 @@
           "type": "string"
         },
         "version": {
-          "description": "The WZDx specification version used to create the data feed",
-          "type": "string"
+          "description": "The WZDx specification version used to create the data feed, in 'major.minor' format",
+          "type": "string",
+          "pattern": "^(0|[1-9][0-9]*)\\.(0|[1-9][0-9]*)$"
```
chuehlien, line 62: Can confirm that the pattern works well. 

Do we want to add an example in the description like what you've done for some other fields? Something like `(e.g. version 2.0 will be represented as '2.0')`. This is optional of course.

spec-content/data-tables/road_event_feed_info.md
```diff
@@ -9,4 +9,4 @@ Field Name | Data Type | Description | Conformance | Notes
 feed_info_id |	ID |	Unique identifier for the data feed. |Required | Primary Key Recommendations on the format of this unique identifier will be made in the future. |
 feed_update_date |	DateTime |	The UTC date and time when the data feed was last updated. |	Required | All date/time formats shall use ISO 8601 Data elements and interchange formats – Information interchange. Example: `2016-11-03T19:37:00Z`
 metadata |	URL |	A link to the metadata file (WZ-metadata.txt). See [Metadata table](/spec-content/data-tables/metadata.md) for a description of the file. |	Optional |
-version |	Text |	The specification version used to create the data feed |	Optional	 |
+version |	Text |	The specification version used to create the data feed in `major.minor` format. |	Optional | Examples: `1.1`, `2.0`
```
chuehlien, line 12: I noticed that you suggested making this field a "Required" field, which I agree is a good idea. Will that happen for v3?

j-d-b: @Mahsa-Ettefagh @chuehlien  I missed this one but would definitely support making it required. The feed is a bit useless if it is not.

Mahsa-Ettefagh: I agree with making this field required. This is the information that data producers can add themselves when setting up the feed and will be helpful to track feed versions in the future.

j-d-b: Done. Note however this change is overwritten by the refactoring of the metadata/feed info which already made "version" required. Good to update here for reference though.

PR #91, opened 2020-05-29
Remove full-spec.md

j-d-b:
`full-spec.md` is a single file that includes all the same information contained in the separate project files within `create-feed` and `feed-content` respectively. However, this created the issue of having to update files in two locations with every update. This is error prone and also less clean as it is clearer to have a single source of truth for the spec (the individual files).

The benefit of a single file containing all the specification content, for download or ease of reference, is still captured via the generated PDF for each spec version available on the project Wiki.


sknick-iastate:
Agreed.  I think we can instead put a combined version in the [wiki](https://github.com/usdot-jpo-ode/jpo-wzdx/wiki) if users prefer to have a single combined file.

davidcraig4300:
I agree with this change.

Issue #66, opened 2020-01-10
Including `lane_edge_reference` on every lane in a road event is confusing


j-d-b:
# Including `lane_edge_reference` on every lane in a road event is confusing

## Background
The `lane_edge_reference` property defines the start side (either *left* or *right*) for identifying the lanes within a road event by `lane_number` index. For example, if the `lane_edge_reference` is *left*, the `lane_number` *0* refers to the leftmost lane.

## Issue
Including the `lane_edge_reference`  is a simple and effective technique for clarifying the often confusing indexing of lanes. `lane_edge_reference` only needs to be defined once for each road event, rather than for every lane in the road event as is currently done. Defining it for every lane:
 
* Is confusing, as you could use *right* for some and *left* for others within the same road event
* Requires more code to parse the feed, as you must check the field for every lane
* Adds unneeded fields and leads to larger files

### Currently (possible scenario)
```json
{
  "road_event": {
    "total_num_lanes": 2,
    "lanes": [
      { 
        "lane_edge_reference": "left",
        "lane_number": 0
      },
      {
        "lane_edge_reference": "right",
        "lane_number": 0
     }
    ]
  }
}
```

*Note the above is pseudocode*

## Solution
Moving the `lane_edge_reference` as a conditional field (required if using "lanes") in the `road_events` table would be a simple solution.

### Example
```json
{
  "road_event": {
    "total_num_lanes": 2,
    "lane_edge_reference": "left",
    "lanes": [
      { 
        "lane_number": 0
      },
      {
        "lane_number": 1
     }
    ]
  }
}
```

*Note the above is pseudocode*

j-d-b:
Along with standardizing the starting lane number in #67, instead of moving `lane_edge_reference` to the road event entity, we could standardize this value and remove the field (i.e. require left or right as part of the spec).

This would solve the original issue and simplify the spec, as it removes an enumerated type as well as a field. Unless the option to specify the direction of count greatly lowers the effort for feed producers (please chime in), I think standardizing the counting edge reference and removing the field is the better approach to resolve this issue.

CraigMoore-Sea:
This makes sense. Keeping this as an option lowers the barrier to entry since both left and right standards are in use. 
After we adopt a more complete standard for defining lanes we can reconsider this, likely not something that will happen in v3.

j-d-b:
@CraigMoore-Sea by "this makes sense", are you referring to moving the `lane_edge_reference` to the road event?

[...]

PR #94, opened 2020-05-29
Clarify lane types and usage of lanes table

### LINKED ISSUE #72: Lane Type enumerated type allows values that don't make sense given the context it is used in
# Simplify the Lane Type enumeration

## Background
Currently, the [Lane Type Enumerated Type](/feed-content/enumerated-types/derived-from-its-standards) allows values that don't make sense given the context it is used. This type is used by the `lane_type` field on the lane table--each entry in the table represents a single lane. Many of the allowed types are plural and leftover from WZDx v1 when this enumeration was used on the activity/road event level. Examples: `middle-two-lanes`, `left-3-lanes`, etc.

## Quick Solution
A quick solution would be to disallow the values that no longer apply. This should have been done for the v2 release but was likely not caught in review.

## Simplifying the Lane Type enumeration
While the quick solution solves the biggest issue with the current enumeration, this is a good opportunity to think about what type of lane information we want to get from the `lane_type` field and thus simplify the enumerations of the Lane Type Enumerated Type.

If we just disallowed the plural/non-applicable values, we'd still have several enumerations which are redundant given information from other fields on the lane object. 

An example of a redundant lane type is `middle-lane`. This value indicates if a lane is the "center most lane where are a total of an odd number of lanes". Since a lane also a `lane_number`  and the road event has `total_num_lanes`, you could already determine if the lane was in the center.

The`lane_number` isn't required, but if a producer could provide the `lane_type` as `middle_lane`, they could also give the `lane_number` for the lane. 

## Conclusion
An update needs to be made to the Lane Type Enumerated Type and I'm hoping this issue thread can start the discussion on what route we should take to do that, as there is potential to simplify the enumeration as part of the change.

### LINKED ISSUE #98: Standardize lane numbering start side
This issue was spawned during discussion about solutions for #66.

There appears to be great interest removing the ability for a producer to choose the side (`left` or `right`) from which lane number counting begins and instead standardizing the side as part of the WZDx specification.

#66 was more specifically resolved in #97, intended for the v3.0 release, without standardizing the counting start side. Thus this issue is intended to be a place to continue the discussion regarding standardizing the start side.

I know @davidcraig4300 has some thoughts on this.

j-d-b:
This PR resolves #72 by updating the [Lane Type Enumerated Type](/feed-content/enumerated-types/derived-from-its-standards/lane-type.md) as well as clarifying the usage of a "lane" and the meaning of lane types. As part of that process, it addresses other lane-related issues, such as ordering. Three sets of sequential changes were made; each depends on the previous. I can refactor the code changes in this PR to address either just **1**, **1** and **2**, or all **1**, **2**, and **3**:

## 1: Clarify usage of lanes table and clean up lane types

The following changes are proposed to cleanup and clarity the ambiguity in v2.0: 

1. Clarify that an entry in the [lanes](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/clarify-lane-types/feed-content/data-tables/lanes.md) table represents a single lane
2. Remove all plural values from the [Lane Type Enumerated Type](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/clarify-lane-types/feed-content/enumerated-types/derived-from-its-standards/lane_type.md), such as `both`, `none`, `left-two-lanes`, `middle-two-lanes`... etc.
3. Rename lane type values `outside` and `inside`, which were ambiguous but intended (by description) for shoulders, to be `left-shoulder` and `right-shoulder`, which then matches the TMDD LaneRoadway enumeration values.
4. Change `middle-lane` description to note it can be used for any lane that is not the right- or left-most lane, not just the "center lane when the total number of lanes is odd", and added a new type `center-lane` to capture the former `middle-lane` functionality. This is more appropriate as the TMDD LaneRoadway has a specific "center lane" type, which is different from `middle-lane`.
5. Note that WZDx lane type `bike-lane` relates to the TMDD LaneRoadway `cycle-lane` (these were not connected before)
6. Clean up the [Lane Type Enumerated Type](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/clarify-lane-types/feed-content/enumerated-types/derived-from-its-standards/lane_type.md) file by moving unused TMDD LaneRoadway types to their own list, to not clutter the table with values used in WZDx
7. Add new lane types `center-left-turn-lane`, `reversible-lane`, and `hov-lane` to cover missing functionality.

Note the spec still doesn't require including any lane information or a lane entry for every lane. The key breaking change is that it disallows having an entry in the "lanes" table represent more than one lane. This change doesn't increase the lane information required to produce the feed output, as if one was using `right-two-lanes` as the `lane_type` for an entry in the "lanes" table, they would just split that into to two entries. 

## 2: Add `order` property to the lane to indicate position/index and leave flexibility in `lane_number`

This PR originally proposed requiring `lane_number` for all lanes, as that was intended to be used as an index to identify the position of the lane on the roadway. Currently (in v2.0) with `lane_number` as optional one must depend on the `lane_type` to determine the location of the lane. From subgroup meeting discussions it became clear that `lane_number` was being used differently by various parties. Thus this PR now proposes adding a new required field to the lane entity, `order`, which should be used to indicate the lane's position in sequence on the roadway. `lane_number` will remain optional and openly defined as it was in v2.0. The addition of `order` greatly simplifies parsing the `lanes` array on the feed as a consumer doesn't need to check the `lane_type` and implement logic to determine the location of a "lane" in this array. 

The changes are specifically as follows:

1. Add new required `order` property to the lane entity. This must be an integer value and a value of `1` must be used to indicate the lane (used generally, can be shoulder, bike lane, sidewalk, etc) closest to the roadway edge indicated by the `lane_edge_reference`. An increase in `1` of the `order` value must represent moving one lane over away from the reference edge.
2. Rename `lane_type` to `type`, `lane_status` to `status`. This is to not cause confusion from the abstract use of the word "lane" (as a "lane" here can be a shoulder or a sidewalk), as well as to give distinction to `lane_number`, which is often used for a regular, driveable lane, and because the prefix is not needed anyways given the entity/object name of "lane".
3. Move the `lane_edge_reference` field from the lane to the road event (see #66)
4. Add new `shoulder` lane type, to be used when the shoulder isn't clearly left or right (e.g. between HOV lane and main roadway), or just in place of `left-shoulder` and `right-shoulder`, as whether it is right or left is able to be determined from the `order` value. A future version of WZDx will likely see the deprecation of all lane types that reference a side.
5. Add new generic `lane` lane type, used when a more specific value is unsuitable (or, like `shoulder`, in place of `right-lane`, `left-lane` etc as the location of the lane is able to be determined by the `order`)

**Note:** the `lane_edge_reference` change is the same as #97, and will override #97 if it is approved.

## 3: Standardize lane counting to begin from the **left** side of the roadway and remove `lane_edge_reference`.
The confusion of allowing the producer to choose the side of the roadway to count from is not worth the minor convenience of not having to translate their numbering when creating a WZDx feed. Having the edge reference consistent in all WZDx feeds limits parsing logic as well as making it easier to consistently talk about lanes.

The **left** side was chosen as this is what is used by:
* Open Street Map (OSM)
* AASHTO NTIMC document, "Traffic Incident Response - Lane Designation Terminology"
* @davidcraig4300
* Everyone who spoke up at subgroup meetings

The changes here are simple:

1. Remove the `lane_edge_reference` field from the lane entity
2. State in the business rules that `lane_number` and `order` counting must begin from the **left** side of the roadway.

**Note:** this resolves #98. Also, this will override the changes to `lane_edge_reference` mentioned in **2.** above, as well as overriding #97.


j-d-b:
Looking into the TMDD LaneRoadway enumeration, it appears that the TMDD values are not dash-delinated. See [here](https://www.ite.org/pub/?id=21873631-03DC-A90B-A0E6-AD04D834AD10). Thus we should update the TMDD references (not the WZDx values) to reflect these.

j-d-b:
@sknick-iastate @davidcraig4300, how about the term `reversible-lane` (inspired by [this page](https://en.wikipedia.org/wiki/Reversible_lane)) instead of `bidirectional` for indication the zipper/suicide lanes or center turn lanes?

j-d-b:
Per today's subgroup meeting, the conformance of the `total_num_lanes` field on the road event needs to be updated to be conditional—required if there are lanes associated with the road event.

j-d-b:
PR description has been completely rehauled, though I have not yet made the changes described in the code.

j-d-b:
Changes implemented for all three parts.

spec-content/data-tables/lanes.md
```diff
@@ -1,18 +1,18 @@
...
-lane_status|Enumeration; Text|Status of the lane for the traveling public|Required|Allowed values: open, closed, shift-left, shift-right, merge-right, merge-left, alternating-one-way. See [Lane Status Enumerated Type](/spec-content/enumerated-types/lane_status.md)
-lane_type|Enumeration; Text|An indication of the type of lane or shoulder|Required|See [Lane Type Enumerated Type](/spec-content/enumerated-types/derived-from-its-standards/lane_type.md)
+Data Name | Data Type | Description | Conformance | Notes
+- | - | - | - | - | -
+lane_id | ID | Identifies the lane_info record | Required | Primary key
+[road_event_id](/spec-content/data-tables/road_events.md) | ID | Identifies the road event to which a lane information is related. | Required | Foreign key
+order | Integer | The position of a lane in sequence on the roadway. This value is used as an index to indicate the order of all WZDx lanes provided for a road event. | Required | A value of `1` must represent the **left-most** lane and an increase in 1 must represent moving a single lane over from the **left**.
```
chuehlien, line 15: Since `lane_number` is "Non-negative Integer", should `order` be explicitly "Positive Integer"?

spec-content/data-tables/lanes.md
```diff
@@ -1,18 +1,18 @@
 # lanes
```
chuehlien, line 1: Should we add a row for the `restrictions` field (formerly `lane_restrictions`) in this table?

j-d-b: The restrictions are stored in their own table (lane_restrictions). The `restrictions` field is formed when generating the feed from the data tables. You are one of many to be slightly confused by the mapping of the relational model (how the data is stored) to the feed structure, becuase it's not clear and the relational model is note ideal here. We are planning on reorganizing the repo and directories to use an object model that directly maps to the feed structure after all these changes are made for v3.0.

spec-content/data-tables/lanes.md
j-d-b: Yes, in fact they should both be positive integer.

spec-content/data-tables/lanes.md
```diff
@@ -1,18 +1,18 @@
...
-lane_edge_reference|Enumeration; Text|The roadside edge from which lanes are assigned numbers.|Conditional: required if  lane_number is not null|Counting begins from the edge of the improved surface. See [Lane Edge Reference Enumerated Type](/spec-content/enumerated-types/lane_edge_reference.md)
-lane_number|Non-negative Integer|The number assigned to a lane (not including shoulders)|Optional|Assigned by counting from right or left edge of the improved surface. Counting begins from the edge indicated in the lane_edge_reference field. Useful for text to voice translation.
-lane_status|Enumeration; Text|Status of the lane for the traveling public|Required|Allowed values: open, closed, shift-left, shift-right, merge-right, merge-left, alternating-one-way. See [Lane Status Enumerated Type](/spec-content/enumerated-types/lane_status.md)
-lane_type|Enumeration; Text|An indication of the type of lane or shoulder|Required|See [Lane Type Enumerated Type](/spec-content/enumerated-types/derived-from-its-standards/lane_type.md)
+Data Name | Data Type | Description | Conformance | Notes
+- | - | - | - | - | -
```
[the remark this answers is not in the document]
j-d-b: fixed, these should be using three dashes `---`; I'm surprised the no-space version renders correctly.

spec-content/data-tables/lanes.md
```diff
@@ -1,18 +1,18 @@
...
+lane_id | ID | Identifies the lane_info record | Required | Primary key
+[road_event_id](/spec-content/data-tables/road_events.md) | ID | Identifies the road event to which a lane information is related. | Required | Foreign key
+order | Integer | The position of a lane in sequence on the roadway. This value is used as an index to indicate the order of all WZDx lanes provided for a road event. | Required | A value of `1` must represent the **left-most** lane and an increase in 1 must represent moving a single lane over from the **left**.
+type | Enumeration; Text | An indication of the type of lane or shoulder | Required | See [Lane Type Enumerated Type](/spec-content/enumerated-types/derived-from-its-standards/lane_type.md)
+status | Enumeration; Text | Status of the lane for the traveling public | Required | See [Lane Status Enumerated Type](/spec-content/enumerated-types/lane_status.md)
+lane_number | Non-negative Integer | The number assigned to a lane | Optional | Assigned by counting from the **left** edge of the improved surface. Useful for text to voice translation.
```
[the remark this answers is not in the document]
j-d-b: Great point about updating the description; done. Both should start at 1 also.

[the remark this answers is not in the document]
chuehlien: Awesome! Remember to update the minimum value for `lane_number` to 1 in the JSON schema file.

sbma44:
Apologies for coming in late. It looks like 2 and 3 propose contrasting lane-counting schemes--one from the edge, one from the left. Is this correct, or am I misunderstanding? If so, could we standardize on a single counting system?

The consensus and prior art around counting from the left makes me think there's probably a good reason for it, but I will note that counting from the centerline allows for more consistency over time even if road geometry changes. One can imagine a basemap getting out of date as a new lane is added, then applying a WZDx alert against the now-stale data. Counting from the centerline would allow for the data about lanes present in both copies to remain correct (and potentially be a signal that an update is necessary).

j-d-b:
> Apologies for coming in late. It looks like 2 and 3 propose contrasting lane-counting schemes--one from the edge, one from the left. Is this correct, or am I misunderstanding?

This is correct. If all three levels of changes are approved as implemented here, the ability to provide the reference edge for counting as discussed in Part 2. will be removed (no `lane_edge_reference` field).

Issue #67, opened 2020-01-10
Change `lane_number` to `lane_index`; clarify/standardize first value


j-d-b:
# Change the lanes table `lane_number` field to `lane_index`; clarify/standardize first value

## Background and Issue
Currently, each lane's location is identified by an index, the `lane_number`. It is not specified what the first index should be (that is 0, 1, or 5 even--there's nothing stopping you!). One could also build the following lanes for a two-lane road and they wouldn't be violating the specification.

```json
{
  "road_event": {
    "lanes": [
      {
        "lane_number": 15
      },
      {
        "lane_number": 3
      }
    ]
  }
}
```

*note the above is pseudocode*

## Solution
Since the number is an index, the field name `lane_index` would be more clear. It should be clarified in the field description what the first/start lane index is, either 0 (I'm for this because indexes usually start at 0) or 1.

kris-codes:
In traffic incident management, the far left (inside) lane is designated as lane 1.  This may be a topic of conversation as DOTs and developers have conversations rolling the spec out.

sergebeaudry:
> In traffic incident management, the far left (inside) lane is designated as lane 1. This may be a topic of conversation as DOTs and developers have conversations rolling the spec out.

Currently it depends from state to state. That is why the specs is currently flexible and feed shall declare the edge reference within the lane_edge_reference field

lynnerandolph:
As each state is going to have to implement something to convert their current representation to this representation, have the spec clearly define how it should look will reduce confusion.

CraigMoore-Sea:
Agree this makes sense. The should be considered index values only for the specific record in question. Thus they don't have meaning in relation to another record for the same road further down the road when the lane configuration may change.

Mahsa-Ettefagh:
Comments received during the [V3 Specification Update Subgroup](https://github.com/usdot-jpo-ode/jpo-wzdx/wiki/WZDx-v3-Specification-Update-Subgroup-Meeting-Notes,-2020-04-29) meeting on 4/29/2020:

- Hua Xiang - Maryland DOT: Which DOT use 0 as the right most lane?
- David Craig @ GM: Maps tend to start with 1 and there is no such thing as lane zero. Use lane_number
- Lynne Randolph: starting at 0 is a programming thing, a feed should start with 1
- Hua Xiang - Maryland DOT: Agree. Support: Lane Numaber, and starts with 1
[...]

PR #97, opened 2020-06-03
Refactor lane edge reference and clarify lane numbering

j-d-b:
This PR resolves issues #66 and #67 by:

1. Removing `lane_edge_reference` from the **lane** entity
2. Adding equivalent `lane_number_start_side` field to the **road event** entity
3. Clarifying that  `lane_number` must start at "1" for the lane closest to the roadside edge indicated by `lane_number_start_side`
4. Clarifying that an increase of `lane_number` by "1" represents moving one lane over away from the direction indicated by `lane_number_start_side`


sknick-iastate:
This looks good to me. A couple of comments based on our last spec group meeting and discussing counting shoulders, bike lanes, etc  in the lane number

- I'm not sure I would interpret that shoulder/non-vehicle lanes should be included in the lane number without the background discussion we have been having recently. Especially since I don't think most agencies include these in the lane numbering currently. It may also be due to us including shoulders in the lanes table.  Should we add some text in the `lane_number` description that says this? It may also be good to say `lane_id` instead of just lane when referring to the numbering as well. 

- Should we add a shoulder with a lane number to the examples (specifically a right shoulder to `road_event_id` 12345 in both the multi-point and linestring example)

rchoyler:
Recommend NOT including shoulders in the lane numbering sequence; unless they are designed to accommodate standard vehicular travel which is equivalent to that of the main lanes (i.e. similar road surface, supports same speed limit, same legality, etc.); such as lanes designated for use only during rush hour periods, HOV/HOT lanes, etc. Perhaps "unimproved shoulders" could be designated as LS (left shoulder) & RS (right shoulder). If numeric values are required, then perhaps use 0 for the left shoulder and 99 for the right shoulder.

j-d-b:
Note #94 completely replaces this. If #94 is not approved/merged, this PR will again be relevant. If #94 if merged, this PR can be closed.

j-d-b:
Closed with #94 merge

PR #133, opened 2020-09-21
Revert "Add business rules to the create feed readme"

j-d-b:
This PR reverts usdot-jpo-ode/jpo-wzdx#116, which was merged without approval.

Specific issues were:
* The "work scenarios" should not be included (update: I see the markdown file was removed, but all the images remain...)
* The mobile work zone business rule needs background info, IMO
* The rule about users adding non-WZDx features is not correct according to the JSON schema, which is the single source of truth, and thus one of them needs to be updated.

There may be others. Let's get review from @chuehlien and @Mahsa-Ettefagh at least before merging this time.

Let's update this PR as needed to not revert all changes (as it does now, by default), but only those that should not have been made/need editing.


j-d-b:
@wesleyalforddot @Mahsa-Ettefagh this is fine now IMO. I created this PR so I can't approve. I've tagged you both as reviewers. Once either of you approves I will merge it. Thanks~

Mahsa-Ettefagh (approved):
Please fixt the first sentence under the Business Rules (delete "are" in the "The following business rules are help assure a standardized and interpretable use of the WZDx specification"). Looks good otherwise.

PR #136, opened 2020-10-13
Correct the ERD + add brief section on updating the ERD

j-d-b:
The ERD incorrectly had the `creation_date` field on the `road_event` and the `restrictions_units` field on the `lane_restriction` in bold to indicate they are required, when they are actually optional.

This PR makes those two items non-bold. In addition, it adds a brief paragraph to the `spec-content` README detailing the process for updating the ERD/the drawio file.


chuehlien (approved):
Thanks for catching this and for adding instructions on how to update the ERD. Looks great to me!

PR #144, opened 2020-11-12
Remove last update line from project root README

j-d-b:
This PR removes the following line from the project root README:

> Last updated 9/3/2020 - WZDx Specification v3.0

The change was made for the following reasons:

- The line is unnecessary—one can readily see via the GitHub web UI when the repo and every specific file was last updated
- The line has not been updated consistently when changes were made
- It is not clear what updates would warrant updating the line
- One can check both the Releases section of the right pane on GitHub and the `RELEASES.md` file and the most current release

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,6 @@
 
 This repository contains the Work Zone Data Exchange (WZDx) Specification.
 
->Last updated 9/3/2020 - WZDx Specification v3.0
-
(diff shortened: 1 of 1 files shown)


chuehlien (approved):
Good catch!

j-d-b:
@chuehlien thank you for the quick review and for the great work on GitHub recently! We are both using the verb-first commit messages and the review feature and the repo looks really clean.

PR #146, opened 2020-12-16
Move road event ID to from RoadEvent "road_event_id" to RoadEventFeature "id"

### LINKED ISSUE #141: Include `id` property on the `RoadEventFeature` to better conform to GeoJSON
## Background
When reviewing the entire spec while refactoring to follow an object model, I came across this quote from [section 3.2 of the GeoJSON specification](https://tools.ietf.org/html/rfc7946#section-3.2):

> If a Feature has a commonly used identifier, that identifier SHOULD be included as a member of the Feature object with the name "id", and the value of this member is either a JSON string or number.

In the WZDx realm, road event's all have IDs, provided on the `road_event_id` field on the [RoadEvent](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/spec-content/objects/RoadEvent.md) object. The `RoadEvent` object is used by the [RoadEventFeature](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/spec-content/objects/RoadEventFeature.md) object, which is an instance of a GeoJSON Feature.

## Issue
The current WZDx specification details GeoJSON features that have an ID, however the ID is not included as an `id` field on the Feature itself and thus does not follow the GeoJSON specification.

## Solution
To better follow the GeoJSON specification, the road event's ID should appear on the `RoadEventFeature` object as an `id` property rather than the `road_event_id` property on the `RoadEvent`. 

Additionally using the `road_event_id` property on the `RoadEvent` should be deprecated (and the property eventually removed).

j-d-b:
This PR resolves #141.

Specifically, the following changes were made:
- Deprecate the `road_event_id` field on the [RoadEvent](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/spec-content/objects/RoadEvent.md) object and change the conformance to optional
- Add a conditional `id` property to [RoadEventFeature](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/spec-content/objects/RoadEventFeature.md) object. This will be the preferred place to provide the ID for a road event. This field is required unless the road event's ID is given by the deprecrated `road_event_id` field on the RoadEvent.
- Update the examples by moving the `road_event_id` to the new `id` property


chuehlien (changes requested):
@j-d-b Good work! Everything looks good but I was not able to get a sample feed with only `road_event_id` (and no `id`) to pass the JSON validator with the JSON schema. I don't know how to fix this yet but commenting here in case you have insights.

j-d-b:
> @j-d-b Good work! Everything looks good but I was not able to get a sample feed with only `road_event_id` (and no `id`) to pass the JSON validator with the JSON schema. I don't know how to fix this yet but commenting here in case you have insights.

Thanks for reviewing. I'll look into it. I may have done the conditional logic incorrectly.

chuehlien (approved):
Can confirm that the json schema can handle both cases. Nice work!

PR #148, opened 2020-12-16
Deprecate the total_num_lanes property on the RoadEvent object

### LINKED ISSUE #140: Ambiguous value of `total_num_lanes`
## Background
Lane numbering, whether something like a shoulder counts as a lane, and related topics have been a subject of much discussion throughout the WZDx spec development. The v3.0 release added the `order` property to the [Lane](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/spec-content/objects/Lane.md) object to identify a lane's position on the roadway rather than using `lane_number` for this as agencies were using `lane_number` differently. Most importantly, they are using the lane number when speaking about lanes (e.g. "lane one is closed"), in the typical case, a shoulder or bike lane (which are "lanes" in WZDx terms) would not be given a lane number and would not be counted in `total_num_lanes` on the [RoadEvent](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/spec-content/objects/RoadEvent.md).

## Problem
It is not clear whether `total_num_lanes` should be the number of standard general travel lanes (traditionally numbered lanes) or all WZDx lanes.

Practically, that means `total_num_lanes` could differ from the length of the `lanes` array on a `RoadEvent`. 

Note if `lanes` are provided, regardless of how `total_num_lanes` is being used, it does not provide any additional information as it can be gathered from the `lanes` array.

## Next steps
It needs to be clarified what the value of `total_num_lanes` should reflect.

j-d-b:
This PR resolves #140. Specifically, the following changes were made:

- Marked the `total_num_lanes` property on the [RoadEvent](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/spec-content/objects/RoadEvent.md) object as deprecated, to be removed in a future version
- Added a new business rule requiring that *if* lane information is provided (i.e. the `lanes` property on the RoadEvent is used), there must be an entry in the `lanes` array for every lane in the road event, not just a subset.


chuehlien (approved):
Looks great!

PR #150, opened 2020-12-16
Add road_names property to replace road_name and road_number

### LINKED ISSUE #95: Best Practices for Road Number
Issue name: Best Practices for Road Number

Road number requires a designator for county, state or interstate. Yet there are no rules for its value.  Either there should be a business rule for interstate, each state and county or a standardized list for every numbered road in the US.

j-d-b:
This PR resolves #95.

~~Specifically, the following changes were made:~~
- ~~Allow an array of strings for `road_name` on the [RoadEvent](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/spec-content/objects/RoadEvent.md) object, for providing multiples names and numbers for a given roadway. The single string value is also still allowed, so this is backwards-compatible. Producers should not feel obligated to include multiple names for a roadway~~
- ~~Mark `road_number` on the [RoadEvent](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/spec-content/objects/RoadEvent.md) object as deprecated and note whatever producers were using for a value here can be included in the `road_name` array~~
- ~~Update examples to make `road_name` an array and move the example `road_number` value to this array, also removing the `road_number` property~~

Edit: the approach has been updated per the below comments.


chuehlien (approved):
Looks great!

daylesworth:
I'd like to note that this makes parsing WZDx harder in statically-typed languages (like Swift and Kotlin) since parsers have to determine whether a string or array is present with custom code. Could we consider a new key named `road_names` to hold an array (that could contain just a single string)? Are there other WZDx properties that can hold multiple data types?

j-d-b:
@daylesworth this is a good point and I agree, it's both cleaner and easier to parse for only one type to be allowed. That way is also consistent with `parents`, `children` on the Relationship object.

I considered that approach when making the PR but went for the multiple allowed types approach to preserve backwards compatibility. After v3.0 the goal was to preserve backwards compatibility where possible.

Perhaps the better approach is to:

1. Add a new optional `road_names` property, which must be an array of strings
2. Deprecate `road_name` and `road_number` properties and make optional

j-d-b:
@chuehlien thank you for reviewing!

j-d-b:
This PR has been updated to the above approach, that is:
- Deprecate `road_name` and `road_number`
- Add new conditional `road_names` property (array of strings), required if `road_name` is not provided
- Change conformance of `road_name` to be conditional, required if `road_names` is not provided

create-feed/schemas/wzdx_vNext_feed.json
```diff
@@ -279,12 +279,23 @@
...
+        },
+        "road_name": {
+          "description": "***DEPRECATED*** Publicly known name of the road on which the event occurs",
+          "type": "string"
         }
       },
+      "anyOf": [
+        { "required": ["road_name"] },
+        { "required": ["road_names"] }
+      ],
```
chuehlien, line 295: Nice work in addressing the conditional requirement here!

chuehlien (approved):
The updated approach looks great! I made some minor changes (see above 2 commits) but think this is good to go if those changes look good to you. Good work!

create-feed/schemas/wzdx_vNext_feed.json
j-d-b: Thanks!

j-d-b:
> I made some minor changes (see above 2 commits) but think this is good to go if those changes look good to you.

Yes thank you for those changes, as usual I'm glad you caught them. I forgot to redo the working when I changed the approach from making `road_name` allow an array to making a new field.

Issue #86, opened 2020-05-07
Local Access Only


CraigMoore-Sea:
**Issue name**: Local Access Only
  Need to communicate that local access only is allowed on certain roadways.
In some cases a roadway is closed to through traffic but is open for local access to businesses or residents.

**Motivation**
The current spec only allows for full closure of a roadway with specific restrictions such as no trucks, no parking, weight restrictions, etc, which can be applied to the entire road event or just specific lanes. The spec also allow for noting the vehicular impacts such as all lanes closed, some lanes closed.

**Proposed changes**
To add a new enumerated type to the Road Restrictions values = Local Access Only

Consideration needs to be given to how such a value would interact with Vehicle Impact and Lane Status values to ensure clarity. This may happen through business rules. For example, would producers use Lane Status = closed and a Local Access Only restriction or would the not Lane Status =  open with a Local Access Only restriction.

CraigMoore-Sea:
Another way to achieve a similar result is to place point closures at the end of a road segment to prevents through routing. One point closure for each travel direction at each ends of the road segment allows full access to the address on the segment but prevents through routing. See the diagram below for an example.

Is this preferable to adding a 'Local Access Only' enumerated value to RoadRestrictions that would be used in the _restrictions_ field of the RoadEvent object?

If the 'Local Access Only' value is used in Restrictions of the RoadEvent object then what value should be used in the _vehicle_impact_ field of the RoadEvent object?

![Local Access Only Example](https://user-images.githubusercontent.com/53792644/101523878-0afc6200-393e-11eb-9e27-b7bf34543426.jpg)

j-d-b:
Implemented in #152 and incorporated in the v3.1 release.

PR #152, opened 2020-12-17
Add local-access-only restriction

CraigMoore-Sea:
This PR resolves Issue #86 

Actions taken by this PR include:

- Add a new enumerated value of `local-access-only` to RoadRestriction

This will be used to denote that the only travel allowed in the extent of the RoadEvent is for direct access to addresses located within the extent of the RoadEvent; no through travel is allowed.


j-d-b:
@CraigMoore-Sea thanks for making the PR! I added the new enum values to the schema and pushed those changes as well.

j-d-b (approved):

PR #160, opened 2021-03-05
Move Direction and LaneType enumerated type Markdown files

j-d-b:
This PR is an organization/clarity refactoring.

The changes implemented move the `Direction` and `LaneType` enumerated type Markdown files from the nested `spec-content/enumerated-types/derived-from-its-standards` directory to more simply `spec-content/enumerated-types` to reside alongside all other WZDx enumerated types. As WZDx has evolved, other standards are considered in all updates to the specification, so it was no longer relevant to isolate two specific enumerated types as linked to ITS standards. LaneType has also deviated far enough from its TMDD roots that it is unnecessary and confusing for it to be in a special sub-directory.

References to other standards were still kept within the files, so the history is not lost.

In addition to moving the files, the content was restructured to match all other enumerated type markdown files.


chuehlien (approved):
Great work!

Mahsa-Ettefagh (approved):
Looks good. Thank you.

PR #158, opened 2021-02-12
Create feed updates

wesleyalforddot:
This pull request adds a documents folder to include final work products and minor changes to readme files to include the self-validation checklist and the the early adopter's guide created by the technical assistance subgroups

PR #161, opened 2021-03-05
Clean up document names and hyperlinks

j-d-b:
This PR is my delayed review of #158.

There is no content change here, just replacing spaces with underscores and consistently capitalizing words in file names and changing the Markdown links to internal to the repo, for clarity, consistency, and so they don't break when the branch they referenced is deleted.


j-d-b:
@wesleyalforddot if you have dates for these, ideally we would append those to the document names. Since they are binary files they can't be tracked by git like the rest of the spec.

chuehlien (approved):
Good work!

PR #166, opened 2021-03-11
Clean up HTML tag in a value of the RoadRestriction enumerated type in the spec

Dunge:
I noticed there was a leftover \<br> tag in the enumerated value table that shouldn't be there. Since it's part of the value field, it would show up as text and not formatted as a new line as it does in the description part.


j-d-b (approved):

j-d-b:
@Dunge thank you

PR #167, opened 2021-03-25
README.MD editorial update and v3.1 Release notes

Mahsa-Ettefagh:
Pull request includes editorial updates to README.md to reference the new version. Draft release notes for v3.1 are updated as well. There are few missing links that still need to be updated but will not require a PR.


RELEASES.md
```diff
@@ -1,7 +1,25 @@
+# WZDx Specification v3.1
+Released April 2021
```
j-d-b, line 2: @Mahsa-Ettefagh @chuehlien This will need to be updated with the exact date after we merge to master and create the v3.1 release tag.

j-d-b:
@Mahsa-Ettefagh thanks for getting this started. I added to RELEASES and made one change I've been wanting to do for a while in the main README, which is putting a summary of WZDx before the repository navigation section, so that visitors don't have to scroll past the directory structure to see what WZDx is about. I adapted what I put in the new intro from the below sections but some is new, so I'd appreciate a review and revision if needed.

RELEASES.md
j-d-b: Or alternatively we could use the `month year` for every release, thoughts on a preference?

natedeshmukhtowery: I agree with a move to just month/year, dropping a specific day.

j-d-b: @natedeshmukhtowery I'll make that change.

eli104: Would there perhaps be a stored location for releases in the future? Such as a download site where releases (or actual files) might be stored? If so then perhaps Year/Month might be easier for sorting newest to oldest versions.

j-d-b: @eli104 what you're looking for exists here https://github.com/usdot-jpo-ode/jpo-wzdx/releases

Mahsa-Ettefagh (commented):
the changes look good to me. Thank you.

Issue #184, opened 2021-07-19
Require 'update_date' property on the RoadEvent object


j-d-b:
## Background

There is an `update_date` property on the [RoadEventFeedInfo]https://github.com/usdot-jpo-ode/wzdx/blob/v3.1/spec-content/objects/RoadEventFeedInfo.md) (WZDx feed-level info), [RoadEventDataSource](https://github.com/usdot-jpo-ode/wzdx/blob/v3.1/spec-content/objects/RoadEventDataSource.md), and [RoadEvent](https://github.com/usdot-jpo-ode/wzdx/blob/v3.1/spec-content/objects/RoadEvent.md) objects.

The property is defined as:

> The UTC date and time when the data feed/data source/road event was last updated.

The property is beneficial for consumers to assess data quality by the recency of the `update_date` value. It can also be used to indicate to the consumer that something has changed and they should update their cached data. 

Currently `update_date` is required on the feed-level RoadEventFeedInfo but optional on both the RoadEventDataSource and RoadEvent.

The feed-level `update_date` is not ample to show the RoadEvent information is recently updated as producers may populate the feed-level `update_date` with every response to a request for the WZDx feed. This would be a fair approach, as it represents when that feed document was generated. It indiciates nothing about the recency of the road event data, though.

## Proposal

I want to propose requiring `update_date` on the RoadEvent object. The same can possibly be done for the RoadEventDataSource as well, though the latter is less clear what the `update_date` represents.

From working with MassDOT to publish their WZDx Feed, the RoadEvent's `update_date` is a core piece of information and is easily populated by the data processing logic. If there is a change to road event information, we update the value of this property. This allows consumers to note if the data is being maintained and is current.

## Next steps

Current or upcoming producers, please comment if the RoadEvent `update_date` would be a burden to provide.

The intention of all changes to WZDx is never to make the spec harder to implement. My thoughts on this is that it is not a difficult piece of information to gather and is useful. If I am incorrect or missing possible negatives of requiring `update_date`, please provide your thoughts.

j-d-b:
Though a simple change, requiring this property **breaks backwards compatibility**, so if there is support for this and it is not a challenge for other producers to supply, getting it in for the fall 2021 v4.0 release would be ideal. Else it will likely be at least another year before the change can be made.

j-d-b:
Co-chairs discussion of this issue made it clear that before `update_date` can be required it needs to be clarified what the date represents and when it should be updated.

[...]

PR #190, opened 2021-08-06
Clarify RoadEventFeedInfo update_date definition

j-d-b:
Following issue #184, this PR updates the description of the [RoadEventFeedInfo](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/RoadEventFeedInfo.md) `update_date` property to clarify that this timestamp should reflect when the GeoJSON file representing the WZDx feed was generated.

The recency of this property depends on if the producer is following a request/response model where the GeoJSON feed file is generated every time a consumer requests the feed, or if the producer is generating the file on an interval or other trigger and making that file available for consumers.

The WZDx specification does not require either method and there are currently producers doing it both ways.


mark-mockett (approved):
@j-d-b The updated description looks good to me. Let's make sure it gets added to the v4.0 schema before release.

sknick-iastate (approved):
Looks good!

j-d-b:
> Let's make sure it gets added to the v4.0 schema before release.
@mark-mockett completed.

Issue #155, opened 2021-01-27
Allow restriction value and units for road event level restrictions


j-d-b:
*This issue relates to #154. I recommend checking out that one as well.*

## Background
Currently, restrictions (see [RoadRestriction](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/spec-content/enumerated-types/RoadRestriction.md) enumerated type) can be provided at the lane level or at the road event level. At the lane level, restrictions are provided by an array of [LaneRestriction](https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/spec-content/objects/LaneRestriction.md) objects, which consistent of a type, value, and units. At the road event level, restrictions are provided by an array of restriction types, the text string from the RoadRestriction enum only. 

## Allowing restriction values at the road event level
In recent discussions regarding using WZDx, either modified or unmodified, for representing bridge clearances, it was noted that it may be helpful to allow providing a value and unit alongside a restriction given at the road event level. For the bridge clearance example, one could then provide a road event restriction for `reduced-height` and specify the height value rather than having to provide the restriction on every lane. This is both a convenience but also a necessity if the producer does not have lane-level information to be able to provide the `lanes` array on the RoadEvent, but wants to give a restriction with a value.

A downside of allowing restriction values at the road event level is that it makes the specification more complex by requiring a new object (or renaming LaneRestriction and its properties) and having another array of objects. The approach in #154 (removing road event level restrictions) is a possible solution to adding complexity. 

On a related note, a few restrictions apply best to only the lane-level, such as `hov-2` and `reduced-width`. Also, reading through the list of restrictions in the RoadRestriction enum, I can imagine some values may be best applied only at the road event level. Perhaps lane restrictions should use a different enumerated type from road event level restrictions. That can also be discussed here.

## Summary
In summary, the hope is to simplify the specification, both making it easier to implement and understand, ideally without losing functionality. Being able to specify restrictions in two places with different functionality (i.e. values only allowed in one place) seems confusing. 

The options up in the air are:
~~1. Require lanes and remove `restrictions` from the RoadEvent object (see #154)~~
2. Make the `LaneRestriction` generic (e.g. Restriction object) and change the `restrictions` property on the RoadEvent to be an array of the new `Restriction` and the `restrictions` property on the Lane to be an array of the new `Restriction` as well.
3. Do nothing

Also we should:
[...]

Issue #178, opened 2021-07-16
Add restriction-notice to the EventType Enumerated Type


DeraldDudley:
# Add New Event Type:  `restriction-notice`

Adding a `restriction-notice` to the [EventType Enumerated Type](https://github.com/usdot-jpo-ode/wzdx/blob/master/spec-content/enumerated-types/EventType.md) enables State DOTs to warn drivers of hazardous restrictions such as low-clearances beneath bridges.

## Benefits
- Helps mitigate loss of life, personal injury, infrastructure damage, lost time and money, unnecessary emissions, and all the other ill effects of bridge and tunnel strikes.
- Fosters in-vehicle warnings about hazards to navigation.
- Establishes the WZDx specification can be leveraged to make transportation safer by extending it beyond its original purpose.
- Extending the specification may draw more agencies into using the Work Zone components of the specification.

## Costs
- Diversifying the specification is off scope.

## Support

- The New York State DOT has committed to developing a prototype feed.  
- Other State DOTs have expressed interests.
- Google and Here have committed to consuming and testing NYSDOT's feed
- The National States Geographic Information Council endorses the project.

## Example
- [An example of the `restriction-notice` EventType.](https://gist.github.com/DeraldDudley/b7d4271750cbeb0a91b2d438862ec729)

j-d-b:
I like this addition. It is simple, backwards compatible, flexible, and allows using the WZDx model to represent any kind of restriction on the roadway.

Note a `restriction-notice` road event can represent a permanent restriction unrelated to road work or a work zone, such as reduced height due to an overhead structure, reduced width on a bridge, etc.

j-d-b:
After the creation of #191 and spec extension subgroup meeting, it was decided that the restriction notice is better addressed in a new feed rather than the WZDxFeed, which should stay focused on work zones and similar or related road events (e.g. detour).

DeraldDudley:
This should be contingent on issue #203

j-d-b:
Completed by adding the [RoadRestrictionFeed](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/RoadRestrictionFeed.md) and [RestrictionRoadEvent](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/RestrictionRoadEvent.md) in #205 (WZDx v4.0).

PR #175, opened 2021-07-08
Deprecate lane_number

j-d-b:
This PR resolves issue #153.

Specifically, it marks the `lane_number` property on the [RoadEvent](https://github.com/usdot-jpo-ode/wzdx/blob/master/spec-content/objects/RoadEvent.md) object as deprecated.

PR #198, opened 2021-08-24
Refactor Restrictions

DeraldDudley:
This PR resolves issue #155.

Changes reflect...
- Making the Restriction object used by the RoadEvent Object
- Changing name of LaneRestriction to Restriction
- ~~Add RoadEvent-RestrictionNotice Feature~~
- Changed RoadRestriction to RetrictionType
- Generalized wording of Restriction Types to not specifically reference the WorkZone Events
- Changed LaneRestrictionUnit to UnitOfMeasurement
- ~~Uploaded Object Model and image to reflect changes (Can copy objects from my file to master file.)~~
- ~~Uploaded RoadEvent-RestrictionNotice Object Model  (Can copy objects from my file to master file.)~~
- ~~Added and example~~


j-d-b:
@DeraldDudley Thanks for getting this started.

I will have more comments as I review each file, but at a high level:

- It seems like you addressed two issues (#178 and #175) with this PR. The restriction notice (#178) is an unrelated change to the restrictions refactoring and should be a separate PR—especially as following the spec extension meeting two weeks ago it was more clear for that to be part of a new feed.
- Following the discussion on the #175 thread and spec update co-chairs discussion, units should have been named UnitOfMeasure.
- ~~The schema needs to be updated~~ Oh I see your note on this! will do

DeraldDudley:
Curses!  I did mix the two.  Do you want me to go back and delete the "restriction notice" content or do you want to reject it?
Thoughts on how to handle the restrictions-notice feed?

j-d-b:
I think removing those changes is fine, I cloned your forked repo and was working on the schema already so I can do it.

spec-content/README.md
```diff
@@ -24,9 +24,10 @@ This section provides a tabular list of all objects used in the WZDx specificati
 Object | Description
 --- | ---
 [Lane](/spec-content/objects/Lane.md) | An individual lane within a road event.
-[LaneRestriction](/spec-content/objects/LaneRestriction.md) | A lane-level restriction, including type and value.
+[Restriction](/spec-content/objects/Restriction.md) | Restriction along a road event, including type and value.
 [Relationship](/spec-content/objects/Relationship.md) | Identification of both sequential and hierarchical relationships between road events and other entities.
 [RoadEvent](/spec-content/objects/RoadEvent.md) | Information that describes where, when, and what activity is taking place along a road segment.
+[RoadEvent-RestrictionNotice](/spec-content/objects/RoadEvent-RestrictionNotice.md) | Information that describes permanent or persistant restrictions along a linear road event.
```
j-d-b, line 30: Even if `restriction-notice` EventType was added, this would still not be needed.

create-feed/schemas/wzdx_v4.0_feed.json
```diff
@@ -271,12 +271,11 @@
           "minimum": 0
         },
         "restrictions": {
-          "description": "Zero or more road restrictions applying to the road event",
+          "description": "A list of zero or more restrictions applying to the road event",
           "type": "array",
           "items": {
-            "$ref": "#/definitions/RoadRestriction"
-          },
-          "uniqueItems": true
```
j-d-b, line 279: This is removed because it doesn't apply to objects

spec-content/enumerated-types/RestrictionType.md
```diff
@@ -0,0 +1,26 @@
+# Restriction Type
```
j-d-b, line 1: Note there is no change to the values of this enumerated type, it is just renamed from RoadRestriction to RestrictionType

spec-content/objects/Restriction.md
```diff
@@ -0,0 +1,15 @@
+# Restriction Object
```
j-d-b, line 1: This object replaces LaneRestriction (effectively a rename, with simpler property names)

j-d-b (approved):
I reviewed, made changes to the @DeraldDudley's v4.0-release-restrictions branch, re-reviewed, and now approve the changes.

I think we need another reviewer, as I originated several of the changes here and have a tendency to miss typos.

j-d-b:
The specification changes made by this PR are as follows:

- Rename **LaneRestrictionUnit** enumerated type to **UnitOfMeasurement** which allows more general use
- Rename the **RoadRestriction** enumerated type to **RestrictionType** which is more semantic
- Rename the **LaneRestriction** object to **Restriction** which allows more general use
- Remove the `lane_restriction_` prefix from the properties of the **Restriction** object
- Change `units` (on **Restriction** object) to be singular `unit`
- Change the `restrictions` array on the **RoadEvent** object to be a list of the **Restriction** object rather than a list of the **RestrictionType** (the key change from this PR: road events can now have restrictions with a value and unit)

And in addition:

- Update the example feeds to represent the specification changes above

Note that the object diagram was not updated which is the team's approach as we can't merge the diagram files so those changes need to happen after the PR is merged to the release branch. @DeraldDudley added new diagram files specific to these changes, but I removed those because we don't want those files merged in. @DeraldDudley if you have time, you can put the generated object diagram image in a comment in this thread, since you've already taken the time to make them. However, they will have to change based on the difference in the proposed changes from what you had originally.

spec-content/objects/Restriction.md
```diff
@@ -0,0 +1,15 @@
...
+The Restriction object describes a restriction on a roadway or lane. This object is used by both the [RoadEvent](/spec-content/objectsRoadEvent.md) and [Lane](/spec-content/objects/Lane.md) objects.
```
chuehlien, line 2: Minor edit for line 2 (format "Restriction" for consistent styling and fix link for RoadEvent)

The `Restriction` object describes a restriction on a roadway or lane. This object is used by both the [RoadEvent](/spec-content/objects/RoadEvent.md) and [Lane](/spec-content/objects/Lane.md) objects.

chuehlien (changes requested):
Nice work! Minor updates needed for `/spec-content/objects/Restriction.md` and references to the old names need to be updated in `/create-feed/README.md` (lines 31 and 87).

spec-content/objects/Restriction.md
j-d-b: Thanks @chuehlien, I made the change.

create-feed/README.md
```diff
@@ -27,8 +27,8 @@ The tables below detail the specificaton's content and describe the data used to
 
 - The [RoadEvent](/spec-content/objects/RoadEvent.md) object describes a work zone event.
 - The [TypeOfWork](/spec-content/objects/TypeOfWork.md) object describes the work taking place along the road.  If applicable, it indicates if the work changes the roadway's architecture.
+- The [Restriction](/spec-content/objects/Restriction.md) object describes a restriction on a roadway or lane.
```
chuehlien, line 30: Thanks for updating this one @j-d-b! Heads up that Line 87 still need to be updated (RoadRestriction -> RestrictionType) for text and link. Thanks!

chuehlien (approved):
Looks great - great job @j-d-b @DeraldDudley !

sknick-iastate (approved):
Great work @DeraldDudley and @j-d-b.  

My only comment is that we should update the `restrictions` description in the [Lane object table](https://github.com/DeraldDudley/wzdx/blob/v4.0-release-restrictions/spec-content/objects/Lane.md#properties) to match either the schema that was modified to "A list of zero or more restrictions specific to the lane" or similar to the road event level restrictions description of "A list of zero or more road restrictions that apply to the roadway segment described by this road event."

j-d-b:
> My only comment is that we should update the restrictions description in the Lane object table to match either the schema that was modified to "A list of zero or more restrictions specific to the lane" or similar to the road event level restrictions description of "A list of zero or more road restrictions that apply to the roadway segment described by this road event."

@sknick-iastate I made the change.

Issue #129, opened 2020-09-02
Improve machine-usability of location_verify_method 


j-d-b:
Currently,`location_verify_method`, on the **road_event_data_source** in v3.0, is a text field with a very general description:

> The method used to verify the accuracy of the location information

While there is an example included, there is too much flexibility in what ends up in the content of this field for it to be useful for machines. It seems, based on the example *"Survey accurate GPS equipment accurate to 0.1 cm"*, that this is intended to be just a human-readable, text description of the method. However, there is valuable data (accuracy!) here which could be parsed by machines if structured differently.

I would like to hear from other data producers what they are using as a location_verify_method, as in the example there are two pieces of separate information: 

1. Method: `GPS`
2. Accuracy: `0.1 cm`

I think spatial accuracy especially would helpful to have as a number (units separate). This could instead be added alongside/instead of "estimated" or "verified" for start/end coords, as currently even "verified" doesn't given that much information, especially when `location_verify_method` is not provided (it is optional).

Curious other's thoughts as it seems there is more information that could be added to the **road_event_data_source** to benefit machines.

davidcraig4300:
Jacob,

I would be interested in know this as well.

What you are asking for 0.1cm or 1mm accurate GNSS is extreme survey grade GNSS and it usually requires a static GNSS receiver for many minutes.   Did you really mean 0.1cm?

My belief is, this level of accuracy is not required at all.   Honestly, I don’t believe 1mm accurate GNSS is usable for virtually anything.

I would be thrilled if we could hold the WZDx event reports at 1 meter accurate GNSS.  1m accurate GNSS can be easily achieved with moving machinery and some type of dual-band differential GNSS (PPP, RTX, PP-RTX, etc.).

We actually are at 1.8m accuracy in our GNSS for our Automated Driving vehicles.  Using single band GNSS.

I think we have to ask, what does increased accuracy buy us?  From 1 meter accuracy to something larger like 10+ meters, we can accurately define individual lanes on the road where the larger accuracy cannot.  The limit would be 1.825 meter accuracy, above this you cannot correctly identify individual lanes (DOT spec’d 3.65m wide lane as a reference), below this you can identify lanes.

Going sub meter I don’t believe gives us any more usable information.  Now, I think this would be fun to debate and get other’s inputs.

Thanks,

Dave

j-d-b:
[...]

PR #199, opened 2021-08-26
Location verify method

sknick-iastate:
This pull request addresses the issues identified in #129.  Initial discussion focused on providing a list of enumerations for the location verify method but after further discussion it was determined that verified location information should be made using a GPS equipped device (i.e. not verified through cameras, site inspection, etc).  It was determined that this would be better achieved by updating the [SpatialVerification](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/enumerated-types/SpatialVerification.md) enumerations to clarify that a GPS equipped device is required for verification and depreciating the `location_verify_method`.

As further justification the `location_verify_method` is at the data source level and wouldn't allow for different verification types for individual road events.  If this is a desired then another attribute should be added at the event level.

So in summary the changes are:
- Deprecate `location_verify_method` in the [RoadEventDataSource](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/RoadEventDataSource.md)
- Redefine the [SpatialVerification](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/enumerated-types/SpatialVerification.md) enumerations to clarify that verified work zone locations should use a GPS enabled device

### DIFF
--- a/create-feed/examples/linestring-examples/comprehensive_linestring_example.geojson
+++ b/create-feed/examples/linestring-examples/comprehensive_linestring_example.geojson
@@ -15,7 +15,6 @@
         "contact_email": "solomon.sourcefeed@testcity1.gov",
         "update_frequency": 300,
         "update_date": "2020-06-18T14:37:31Z",
-        "location_verify_method": "GPS",
         "location_method": "channel-device-method",
         "lrs_type": "milemarkers",
         "lrs_url": "https://gis.iowadot.gov/rams/rest/services/lrs"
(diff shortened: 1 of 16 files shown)


j-d-b:
@sknick-iastate let me know what you think of my above approach.

sknick-iastate:
@j-d-b sorry for the delayed response.  I like your approach for changing the verification to a boolean.  I'd like to get feedback at our next specification update meeting then can make the change.

j-d-b (changes requested):
See https://github.com/usdot-jpo-ode/wzdx/pull/199#issuecomment-907206410

PR #217, opened 2021-12-09
Allow decimals for reduced speed limit KPH values

mark-mockett:
This patch allows the renamed `reduced_speed_limit_kph` property to take numbers instead of integers, allowing more precise values for speed limits converted to km/h from mph. This minor change is backward compatible with the previously approved change and only requires approval of the WZDWG co-chairs.


natedeshmukhtowery (approved):

sknick-iastate (approved):

dxpack:
This type change should also be applied to the average_speed_kph properties of the the TrafficSensor and TrafficSensorLaneData objects. There may be additional properties in the spec that should be updated as well, where a speed measurement type is Integer.

j-d-b:
@dxpack thanks, added a new issue for it in #219.

PR #227, opened 2021-12-17
Add additional values for MarkedLocationType enumeration

dxpack:
Location markers in a work zone (or road event) may not be exclusive to the start or end positions, or may be indeterminate in terms of order within a series of location markers. Adding the following additional values for MarkedLocationType:

- road-event-intermediate - A middle point of a road event
- road-event-undetermined - A point in a road event of undetermined order
- work-zone-intermediate - A middle point of a work zone
- work-zone-undetermined - A point in a road event of undetermined order

### DIFF
--- a/schemas/4.0/SwzDeviceFeed.json
+++ b/schemas/4.0/SwzDeviceFeed.json
@@ -323,7 +323,7 @@
             "function"
           ]
         }
-      ]    
+      ]
     },
     "HybridSign": {
       "title": "Hybrid Sign Field Device",
@@ -576,9 +576,13 @@
         "lane-closure",
         "temporary-traffic-signal",
         "road-event-start",
+        "road-event-intermediate",
         "road-event-end",
+        "road-event-undetermined",
         "work-zone-start",
-        "work-zone-end"
+        "work-zone-intermediate",
+        "work-zone-end",
+        "work-zone-undetermined"
       ]
     }
   }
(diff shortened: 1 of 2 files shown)


j-d-b:
What is the utility of these values?

dxpack:
Yes, in the case of the _-intermediate_ values. In the WZDxFeed, the geometry is set to either LineString or MultiPoint, with the former being preferred. A LineString is comprised of **at least** 2 points to describe "the entire road event path". MultiPoint is to be used "when only the start and end coordinates are known". As is (v4.0) the spec for WZDxFeed provides for (and prefers) the delineation of a work zone to include locations intermediate to the start and end points, but as the SwzDeviceFeed does not provide a means to ascribe a LocationMarker device as intermediate, those devices will be excluded from SwzDeviceFeed.

As for _-undetermined_ values, there is also the need to include LocationMarker devices that are components of a work zone but are not part of the delineation of a work zone. For example, a GPS tracking device (LocationMarker) attached to moving equipment that is being used in the work zone, or a LocationMarker attached to a static message sign (or other unconnected/non-electronic device) placed far ahead of the work zone where the static message sign is used to notify upstream vehicles of the upcoming work zone. Although organizationally both are component devices of the work zone, as is (v4.0) these would be excluded from SwzDeviceFeed because there is no applicable MarkedLocationType.

j-d-b:
Superseded by #311.

PR #249, opened 2022-02-23
V4.0 Documents Page update

mollybehan:
- Created a README.md table on the documents page to help clarify what resources are available.
- Added several documents collected by the TA Co-chairs to serve as references for policy language and data security.


documents/README.md
```diff
@@ -0,0 +1,12 @@
+| Document Title | Publisher | Description |
```
j-d-b, line 1: I think you should have line 1 be a header, for example:

```
# Documents
```

Then the table below. Otherwise looks good!

mollybehan: Thanks for the feedback, I've added the header.

documents/README.md
```diff
@@ -0,0 +1,13 @@
+# Documents
+| Document Title | Publisher | Description |
```
j-d-b, line 2: FYI this table markdown can be simpler. You don't need the `|` on the outside of each row, nor as many `------`. It keeps the file smaller and easier to read when not rendered. It is also then consistent with how we represent tables in other markdown files.

For example:

```
Document Title | Publisher | Description
--- | --- | ---
this | row | content
```

j-d-b (approved):
Thanks for making the changes. Approved: you can merge the PR any time.

PR #267, opened 2022-03-20
Allow lane-level traffic sensor data without a defined road event

### LINKED ISSUE #225: Lane level traffic sensor data restrictions
Issue name: “Lane level traffic sensor data restrictions”
 ## Summary
The [TrafficSensor Object](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/TrafficSensor.md) allows for summary data to be collected as well as individual lane data.  The restriction on the lane-level data is that it can only be provided if the data producer knows the road event it is assigned to.  See the note below in the [TrafficSensorLaneData Object](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/TrafficSensorLaneData.md):

> Note this structure allows a single [TrafficSensor](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/TrafficSensor.md) to provide data across lanes on multiple road events. It does not allow for lane data to be provided when a road event is not known.

This seems overly restrictive at this point. I think it should be strongly recommended to provide the `road_event_id` but I would prefer to receive the lane level data even if the data producer wasn't aware of what road event it is assigned to.  I can also see situations where certain lanes are not associated to a work zone (i.e. ramps or opposite direction of travel) that the sensor is collecting data for that would be missed in the current implementation.

 ## Proposed changes
Recommend making the `road_event_id` optional in the [TrafficSensorLaneData Object](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/TrafficSensorLaneData.md) and remove notes about needing an associated road event for lane level data.

j-d-b:
This PR resolves #225, specifically:

- Change the conformance of the `road_event_id` property on the [TrafficSensorLaneData]() object from Required to Optional.

This change allows a producer to provide lane-level traffic data without needing the ID of a defined road event.


j-d-b:
This PR is in draft state until #265 is merged.

mollybehan (approved):

j-d-b:
Description changes  to be made per co-chairs discussion 2022-08-04

TrafficSensorLaneData:

- Update `lane_order` description:
    - First sentence: change "road event" to "roadway";
    - Second sentence: change to "If the `road_event_id` is provided the value of this property corresponds to the associated road event's [Lane](/spec-content/objects/Lane.md)'s `order` property."

TrafficSensor:

- Remove "road event" language from description

mark-mockett (approved):

PR #273, opened 2022-03-21
Add road_direction property to FieldDeviceCoreDetails

### LINKED ISSUE #250: Add direction to FieldDeviceCoreDetails
## Background

The [RoadEventCoreDetails](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/RoadEventCoreDetails.md), used by all WZDx road events, includes a required `direction` property which is defined as:

> The digitization direction of the road that is impacted by the event. This value is based on the standard naming for US roadways and indicates the direction of the traffic flow regardless of the real heading angle.

It's value is restricted to the [Direction enumerated type](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/enumerated-types/Direction.md).

This property is required as a single road event can only describe an impact to one direction of travel. `direction` allows a WZDx consumer to know which direction of the road the road event is on.

## Issue

For field devices, there is no `direction` property. That means a consumer cannot tell what direction of travel the field device applies to. That is problematic in several cases, such as when a consumer is using the location of a device in a [SwzDeviceFeed](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/SwzDeviceFeed.md) to generate a road event. 

*This issue was discussed in https://github.com/usdot-jpo-ode/wzdx/discussions/248*

## Proposed Solution

Adding `direction` to the [FieldDeviceCoreDetails](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/FieldDeviceCoreDetails.md) would resolve the issue.  Perhaps, as discussed in #229, the property should be named `road_direction` or `direction_of_travel` for clarity (as it's the direction of the road the device is on rather than the direction of the device).

## Next Steps

- Should `direction` on the FieldDeviceCoreDetails be required or optional?
- Should a single device be able to apply to multiple directions of travel?

j-d-b:
This PR resolves #250 by adding a `road_direction` property to the [FieldDeviceCoreDetails](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/FieldDeviceCoreDetails.md) object. Specifically, the following changes were made:

- Add optional `road_direction` property to the FieldDeviceCoreDetails object.
- Move the "Direction" JSON Schema type definition to it's own JSON file and generalize its description.
- Reference the moved "Direction" type from SwzDeviceFeed.json and RoadEventFeature.json.

As more enumerated types are shared between feeds, we can consider further organization of the JSON schema files.


mollybehan (approved):

sknick-iastate:
The only minor change I see is that the FieldDeviceCoreDetails needs to be added to the used by section of the [Direction enumerations](https://github.com/usdot-jpo-ode/wzdx/blob/feature/device-road-direction/spec-content/enumerated-types/Direction.md).  Everything else looks good.

eli104:
I apologize if this comment is late to the discussion, and/or if it's already been discussed elsewhere: 
I notice on the Direction Enumeration that the values are based on TMDD link-alignment Enumeration... which itself also notes "inner-loop" and "outer-loop" to help refine the direction of the road flow. 

Are these additional attrobutes something that have, or should be, considered? Perhaps under Direction or perhaps in the RoadEventCoreDetails (further enhancing the requirements of the "road_name")? 

Reasons: 
- A location (automatically from a device) may not be accurate when inner/outer roadways are nearly parallel, or when other road beds provide the same geographic location (such as upper and lower bridge decks, e.g. the George Washington Bridge and others in NY/NJ).  
- The same may be true for a manually entered location, meaning the direction alone may not provide clarity in if the "inner/outer" or "upper/lower" roadway is not specified.

Again my apologies if this has already been discussed (or if the discussion needs to move somewhere else... so be it).

j-d-b:
@eli104 see #324—the discussion can be continued there.

j-d-b:
> The only minor change I see is that the FieldDeviceCoreDetails needs to be added to the used by section of the [Direction enumerations](https://github.com/usdot-jpo-ode/wzdx/blob/feature/device-road-direction/spec-content/enumerated-types/Direction.md).

Thank you, fixed and pushed.

sknick-iastate (approved):

PR #278, opened 2022-03-22
Recommend UUIDs for identifiers and add "name" to RoadEventCoreDetails

### LINKED ISSUE #207: Use UUIDs for feature (RoadEventFeature) IDs
A Universally Unique IDentifer (UUID), also known as GUID, defined in [RFC 4122](https://datatracker.ietf.org/doc/html/rfc4122), is a 128-bit ID that can guarantee uniqueness across space and time. 

Here's an example of a UUID: `6ba7b810-9dad-11d1-80b4-00c04fd430c8`

Currently, WZDx [RoadEventFeature](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/RoadEventFeature.md) `id` is *just a string*. It is likely that there will be collisions between road event IDs from different producers which makes it difficult to aggregate feeds and to relate objects which reference a road event ID (such as within a Relationship](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/Relationship.md), e.g. connecting detours to work zones) to the road event they refer to. Within a single data source there is no issue, but across feeds from multiple producers the ID of the road event alone is not ample to reference a road event. It would be way easier if it was, and a UUID would enable that.

From my experience working with MassDOT and Ver-Mac, this change would not be burdensome as the content of a RoadEventFeature ID does not have meaning and by its description in the WZDx specification is not intended to. Thus the requirement of a UUID would just require a change in how IDs are generated. There are many libraries for generating UUIDs, so this is trivial.

Currently, the only feature in WZDx with an identifier is the RoadEventFeature. However, active PRs seek to add a FieldDeviceFeature (see #195), and there has been discussion about a RoadFeature. I think all feature IDs should be UUIDs which would greatly facilitate referencing features without the context of the original feed or data source they came from, historical reporting and data warehousing, and relating features.

j-d-b:
This PR resolves #207. Specifically, the following changes are proposed:

- Add language recommending using a UUID to the [RoadEventFeature](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/RoadEventFeature.md) `id` property and the [FeedDataSource](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/FeedDataSource.md) `data_source_id` property.
- Add a new optional `name` property to the [RoadEventCoreDetails](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/RoadEventCoreDetails.md)
- Update feed examples to use UUIDs and the `name` property.


mollybehan (approved):

mark-mockett (approved):

examples/RoadRestrictionFeed/bridge_height_restriction_linestring_example.geojson
```diff
@@ -166,7 +168,7 @@
       }
     },
     {
-      "id": "Bridge3",
+      "id": "11d150f3-e08e-43e9-95d8-e185cb62c93f",
```
sknick-iastate, line 171: Bridge2 and Bridge3 have the same UUID

sknick-iastate (approved):

jlarsonOmahaNE (approved):

PR #281, opened 2022-03-23
Use boolean properties for WorkZoneRoadEvent position verification

### LINKED ISSUE #213: Refactor RoadEvent beginning and ending accuracy
This issue is regarding refactoring the `beginning_accuracy` and `ending_accuracy` properties on the [RoadEvent](/spec-content/objects/RoadEvent.md) for clarity.

I originally proposed this idea in https://github.com/usdot-jpo-ode/wzdx/pull/199#issuecomment-907206410 and moved it to and issue as it is separate enough from #199 to be discussed separately.

The proposed changes simplify the specification without losing functionality (which is my goal—make WZDx easy to use and understand while still being powerful).

## Proposed Changes
- Deprecate `location_verify_method` (which is proposed in #199)
- Remove the [SpatialVerification](/spec-content/enumerated-types/SpatialVerification.md) enumerated type
- Remove the `beginning_accuracy` and `ending_accuracy` properties (which use the SpatialVerification enumerated type) from the [RoadEvent](/spec-content/objects/RoadEvent.md) object.
- Add new properties `is_begin_location_verified` and `is_end_location_verified` to the RoadEvent which have boolean values indicating whether or not the begin/end is verified (if it's not verified, it's estimated, so we don't need an enumerated type if there are only two options).

### Pros
- Simpler and requiring less documentation (the new WZDx user doesn't have to reference an enumerated type and see it only has two values)
- The name `is_begin_location_verified` is more clear what the property is describing than `beginning_accuracy`, which doesn't indicate by name that it's for the location (we could name it `is_start_point_verified` to be more clear, because that's what the property is specifically describing).
- Makes it more clear what the business rule for the use of `is_begin_location_verified` could be:

> `is_begin_location_verified` and `is_end_location_verified` must only be `true` when the location is based on actual reported data from a GPS equipped device showing the measured location of the work zone

### Cons
- If we wanted to add more options than just `estimated` and `verified` for describing the accuracy of the start/end
- Breaking change with no functional value.

j-d-b:
This PR resolves #213. Specifically, it proposes the following changes:

- Deprecate the `beginning_accuracy` and `ending_accuracy` properties on the [WorkZoneRoadEvent](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/WorkZoneRoadEvent.md).
- Add deprecation note to [SpatialVerification](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/enumerated-types/SpatialVerification.md) enumerated type.
- Add new boolean properties `is_start_position_verified` and `is_end_position_verified` to the WorkZoneRoadEvent to indicate if the work zone start and end position (first and last geometric coordinate pair, respectively) are based on actual reported data from a GPS-equipped device.
- Require either `beginning_accuracy` or `is_start_position_verified`.
- Require either`ending_accuracy` or `is_end_position_verified`.
- Update example feeds to use `is_start_position_verified` and `is_end_position_verified`.


mollybehan (approved):

schemas/4.1/RoadEventFeature.json
```diff
@@ -150,18 +152,38 @@
...
-            "beginning_accuracy",
-            "ending_accuracy",
             "start_date",
             "end_date",
             "start_date_accuracy",
             "end_date_accuracy",
             "vehicle_impact",
             "location_method"
+          ],
+          "oneOf": [
```
j-d-b, line 174: This should be "anyOf" to allow feeds to contain both the old and new versions, for flexibility before the deprecated properties are removed.

sknick-iastate (approved):

Issue #215, opened 2021-10-21
Representing bidirectional passing lanes


j-d-b:
This issue was created in response to the comment posted by @eli104 in https://github.com/usdot-jpo-ode/wzdx/discussions/131#discussioncomment-1508749_.

Looking over the WZDx representation of scenarios 3 and 10 in #131, I am concerned about expressing a passing lane that is primarily used by traffic in the opposite direction as `left-lane`, no different from any other left lane (in WZDx v4.0, this lane would be given the type `general`). I think this is too unclear and potentially dangerous.

## Example of the issue
See the following example roadway. `--  --  --  --` represents a dashed yellow line, separating opposing directions or travel and allowing passing from either direction. Due to road work (shoulder closure), using the opposing lane for passing is no longer allowed:

```
********************************************************** Road Edge
                    Shoulder
----------------------------------------------------------
                    <--------
--  --  --  --  --  --  --  --  --  --  --  --  --  --  --  
                    --------->
----------------------------------------------------------
                 // Taper              Shoulder (closed)
********************************************************** Road Edge
```

The WZDx v4.0 representation to be consistent with the examples in #131 would be:

### Road Event 1 (eastbound)
```
"lanes": [
  {
    "order": 1,
    "type": "general",
    "status": "closed"
  },
  {
    "order": 2,
    "type": "general",
    "status": "open"
  },
  {
    "order": 3,
    "type": "shoulder",
    "status": "closed"
  }
]
```

### Road Event 2 (westbound)
```
"lanes": [
  {
    "order": 1,
    "type": "general",
    "status": "closed"
  },
  {
    "order": 2,
    "type": "general",
    "status": "open"
  },
  {
    "order": 3,
    "type": "shoulder",
    "status": "open"
  }
]
```

Based on this representation, there is no distinction between the passing lane and the primary travel lane. They both have the lane type `general` which seems confusing.

## Potential Solution

There are two parts to a potential solution:

1. Not represent the passing lane with a [Lane](/spec-content/objects/Lane.md). This alone loses the ability to describe the passing lane.
2. Add a new value to the [RoadRestriction](/spec-content/enumerated-types/RoadRestriction.md) named `no-passing` which would allow representing that the passing lane is no longer available for use. The WZDx representation would then be:

### Road Event 1 (eastbound)
```
"lanes": [
  {
    "order": 1,
    "type": "general",
    "status": "open",
    "restrictions": [{ "type": "no-passing" }]
  },
  {
[...]

PR #306, opened 2022-05-23
Adding no-passing to RestrictionType

jlarsonOmahaNE:
Per Issue #215, this adds `no-passing` as a value to the [RestrictionType](https://github.com/usdot-jpo-ode/wzdx/blob/feature/no-passing-restriction/spec-content/enumerated-types/RestrictionType.md#restrictiontype-enumerated-type) enumerated type.

### DIFF
--- a/spec-content/enumerated-types/RestrictionType.md
+++ b/spec-content/enumerated-types/RestrictionType.md
@@ -18,6 +18,7 @@ Value | Description
 `gross-weight-limit` | Vehicle gross-weight-limit restrictions reduced along the segment being described.
 `towing-prohibited` | Towing prohibited along the segment being described.
 `permitted-oversize-loads-prohibited` | “Permitted oversize loads” prohibited along the segment being described; this applies to annual oversize load permits.
+`no-passing` | The passing lane is no longer available for use.
 
 ## Used By
 Property | Object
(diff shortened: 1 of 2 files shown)


mark-mockett (approved):

spec-content/enumerated-types/RestrictionType.md
```diff
@@ -18,8 +18,9 @@ Value | Description
 `gross-weight-limit` | Vehicle gross-weight-limit restrictions reduced along the segment being described.
 `towing-prohibited` | Towing prohibited along the segment being described.
 `permitted-oversize-loads-prohibited` | “Permitted oversize loads” prohibited along the segment being described; this applies to annual oversize load permits.
+`no-passing` | The passing lane is no longer available for use.
```
j-d-b, line 21: It may be more clear to reword the description to clarify that the lane given the type of "no-passing" isn't the passing lane, it's the lane that you can't pass from starting in. Perhaps:

> Passing from this lane is no longer available.

Let me know if you agree or if you have another suggestion.

Let me know if that makes sense, or if you have suggestions to modify the wording further

natedeshmukhtowery:
Coming in late - is it passing “from” or passing “in”? I might also use “permitted” rather than “available”

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

spec-content/enumerated-types/RestrictionType.md
jlarsonOmahaNE: What about: Passing from this lane is prohibited.

j-d-b: That sounds good to me, but see @natedeshmukhtowery's comment about "from" vs "in". I'm not sure which is the right word.

jlarsonOmahaNE: @j-d-b I did see Nate's comment. After looking at the issue, it looked more like a "from" due to the lanes being in the opposite direction of travel.

I think his comment of "in" would come into view if the direction of travel was the same and passing was restricted. So maybe we should account for both?

j-d-b (approved):

sknick-iastate (approved):

mark-mockett (approved):

Issue #238, opened 2022-01-21
Add new options to FlashingBeaconFunction enumerated type


Dunge:
## Summary
The `SwzDeviceFeed` `FlashingBeacon` entity have a list of [FlashingBeaconFunction](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/enumerated-types/FlashingBeaconFunction.md) that is currently a bit limited.

 ## Proposed changes
I would see a value to add a few fields, like meteorological events. Suggestions could include `low-visibility`, `high-winds`, `overheight-detection`, `railway-crossing`, `pedestrian-crossing`, and more importantly `other` or `unknown`.

j-d-b:
I think that most of these are great additions. The original set was due to the input we had at the time of the WZDx v4.0 development cycle, where these were not brought up. 

The one I am hesitant on with is `unknown`, because I am not sure the scenario where that would be used. @Dunge can you explain your though processes behind that one?

Dunge:
Well in our system, the data set doesn't have a function property on the flashing beacon entity, we only know that it is a flashing beacon and its status. A human can deduce its function based on the owner, name, location and project it is on, but not something that can be filled from code. Even if we do add a similar property in our model, you can be certain most users won't take the time to fill it out when creating the device entity. 

So in this situation, would you rather not have the device present in the feed at all if we are unable to specify its function?

In any case, just `other` would fill a gap. For now all of our flashing beacons serve as some of the new values I suggested, and even if I had the property for in my system to differentiy them apart, I currently can't add them in the feed because they aren't doing one of the function decided in the original list.

j-d-b:
Got it. It is a problem that you can't add flashing beacons that aren't one of the current FlashBeaconFunction enumerated type values.

I agree that `other` should be there and understand the user of `unknown`. It is better that the information that there is a flashing beacon at a precise location can be disseminated than to know what it is—as you noted, the consumer may be able to tell from other properties.

Dunge:
Also note that setting the `function` property of the object as optional would serve the same purpose. It's pretty much the same reason as why any other field can be optional.

sergebeaudry:
Hi regarding this expressed by @j-d-b 
   
> I agree that other should be there and understand the user of unknown. The core information is that there is a flashing beacon at a precise location; having unknown lowers the effort to express that.

[...]

PR #308, opened 2022-05-25
Allow specifying the sign text for a Flashing Beacon

### LINKED ISSUE #295: Add Static Text for Flashing Beacons
Add Static Text for Flashing Beacons: [New feature]
## Summary
Similar to #275 we need the same required field for the associated static text related to any Flashing Beacons to know what the reason/purpose is for the beacons flashing. What are they telling the motoring public? I know to make this a required field will have to wait until the next cycle for V5.0 of the spec/feed, but for this and #275 we should add them now as optional fields and make them required in the next cycle.
 ## Motivation
As of right now there is now way to know exactly what is being conveyed to the motorists, beyond a general function. The specific text will add this clarity.
## Proposed changes
I propose to make the associated static text an optional feed now and when v5.0 is accommodated in a future cycle, we then make it permanent, since we cannot not due to breaking backwards compatibility.

j-d-b:
This PR resolves #295 by adding an optional `sign_text` property to the [FlashingBeacon](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/FlashingBeacon.md) object which allows a producer to provide the text on the sign the beacon is mounted on, if applicable.

### DIFF
--- a/schemas/4.1/SwzDeviceFeed.json
+++ b/schemas/4.1/SwzDeviceFeed.json
@@ -316,6 +316,10 @@
             "is_flashing": {
               "type": "boolean",
               "description": "A yes/no value indicating if the flashing beacon is currently in use and flashing"
+            },
+            "sign_text": {
+              "type": "string",
+              "description": "The text on the sign the beacon is mounted on, if applicable"
             }
           },
           "required": [
--- a/spec-content/objects/FlashingBeacon.md
+++ b/spec-content/objects/FlashingBeacon.md
@@ -9,6 +9,7 @@ Name | Type | Description | Conformance | Notes
 `core_details` | [FieldDeviceCoreDetails](/spec-content/objects/FieldDeviceCoreDetails.md) | The core details of the field device that are shared by all types of field devices, not specific to flashing beacons. | Required | This property appears on all field devices.
 `function` | [FlashingBeaconFunction](/spec-content/enumerated-types/FlashingBeaconFunction.md) | Describes the function or purpose of the flashing beacon, i.e. what it is being used to indicate. | Required |
 `is_flashing` | Boolean | A yes/no value indicating if the flashing beacon is currently in use and flashing. | Optional | The `is_flashing` property is optional as it should not be provided if the producer does not know if the beacon is flashing (e.g. if it's in error state or similar).
+`sign_text` | String | The text on the sign the beacon is mounted on, if applicable. | Optional |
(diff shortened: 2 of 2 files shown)


RHolt-RIDOT:
This added option seems safe for now thank you.  But I have just moments ago added some input to Issue #295 that may pertain to this, and so below I repeat that input in case/hopes it helps with moving forward on this Issue.

-----

Many others have already well noted/described (e.g., via Issue #238) that there are many, many, many possible use cases for Flashing Beacons (FBs) in work zones, and while at first glance I was thinking @tfoster-vm and others have been 100% spot-on above in noting that that spec/feed ideally should accommodate/indicate the reason/purpose for (function of) the FB (i.e., what are they telling the road users?), as I think on it further, it may be helpful to remember the following:  per current MUTCD an FB is almost always SUPPLEMENTING an appropriate warning or regulatory sign, marker, or other TCD (_only exception I can think of is for an Intersection Control Beacon (ICB) where a Yellow FB need not be supplemented by anything else, and in vast vast vast majority of Work Zones there will NOT be any ICBs_).  So I suggest considering whether it makes sense to not focus on the FBs but instead focus on the "greater TCD" or system that each FB is associated with.  Not that anyone wants to see another list, but a 'high-end' generic list of such system functionality that I assume may be most commonly used in Work Zones in near future could be:

- Regulatory Speed Limit In Effect or Active (Regulatory)
- Advisory Speed Limit In Effect or Active (Warning Only)
- Warning, Road Worker(s) Ahead / Be Prepared to Stop or Slow As/If Necessary
- Warning, Traffic Queue or Congestion Ahead / Be Prepared to Stop or Slow As/If Necessary
- Warning, Adverse Weather Condition Detected / Be Prepared to Stop or Slow As/If Necessary
- Warning, New or Changed Traffic Pattern Ahead
- Warning (and/or Regulatory?), Over-Sized Vehicle Detected

I may certainly be missing a few from the above.

If more specific static text to specifically indicate what the TCD is telling road users is needed, then I can understand and am fully open to as proposed above by @tfoster-vm.  I just hope (from my DOT perspective) that, given there are so many, many possible messages on TTCDs that FBs can be used to supplement in a Work Zone, there won't be any undue burden on data providers/contractors (all the more support for making things "optional" at least for now).

sknick-iastate (approved):

mark-mockett (approved):

PR #313, opened 2022-05-26
Enable the Device Feed to describe temporary traffic signals

### LINKED ISSUE #230: Expand connected temporary traffic signals in SwzDevice
Issue name: “Expand connected temporary traffic signals in SwzDevice”
 ## Summary
Currently the SwzDevice feed allows for temporary traffic signals under the field device type of [location marker](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/LocationMarker.md).  The location marker device type is currently designed to only describe the location marker type/location but doesn't allow any additional attributes associated with the device.  In Iowa, there is a desire for an additional attribute on the `mode` of the connected temporary traffic signals that we would like to incorporate into the SwzDevice feed.  Providing the mode allows for a better understanding of the impacts the signal has on traffic as well as ensuring the device is operating as desired.

 ## Proposed changes
Iowa was working on creating a protocol for connected temporary traffic signal and originally put together the following enumerations for the `mode`:
- No communication available to the temporary signal set 
- Powered on, but not displaying
- Using a timed cycle 
- Using a manual trigger
- Using some other external trigger
- Disabled due to an error

I feel there are multiple options for adding a `mode` for the temporary traffic signals but looking for feedback from others on the best implementation that aligns with the SwzDevice feed.  Below is a list of some potential options

- **Option 1**: Simply add a `mode` in the [MarkedLocations Object]( https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/MarkedLocation.md) that is optional and only used when the `MarkedLocationType` is `temporary-traffic-signal`.  The problem with this approach is that the mode is not relevant to most of the marked locations.
- **Option 2**: Create a new device type object for TrafficSignal similar to the [Arrowboard Object]( https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/ArrowBoard.md) and [FlashingBeacon Object]( https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/FlashingBeacon.md).  The TrafficSignal object would contain the `core_details` and `mode`. I would propose making this a required field as well.  This approach would allow additional attributes to be added specific to traffic signals if needed in the future.

j-d-b:
This PR resolves #230. 

Currently, in WZDx v4.0, a connected temporary traffic signal device can be represented using a [LocationMarker](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/LocationMarker.md), however, no information can be provided about it. 

This PR creates a new TrafficSignal field device that allows representing a traffic signal in a WZDx DeviceFeed as its own object with specific properties that apply to it.

The follow changes are proposed:

- Deprecate the `temporary-traffic-signal` value on the [MarkedLocationType](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/enumerated-types/MarkedLocationType.md) enumerated type.
- Add TrafficSignal object with `core_details` and `mode` properties.
- Add TrafficSignalMode enumerated type, used by the TrafficSignal `mode` property.
- Add "See object" references to [FieldDeviceType](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/enumerated-types/FieldDeviceType.md) value descriptions.


RHolt-RIDOT:
thanks for suggesting / advancing this.  Note, current MUTCD defines a Flashing Beacon as a type of Traffic Signal, so suggest checking/confirming that there won't be any issues or confusion with the terminology being used (either term above).

sknick-iastate (approved):

mark-mockett (approved):

PR #335, opened 2022-08-17
Release WZDx v4.1

mark-mockett:
This PR will officially release WZDx v4.1. This notes section will be updated with a changelog in the next couple of days

PR #336, opened 2022-08-17
Update documentation and make editorial changes for the v4.1 release

j-d-b:
This PR proposes changes needed to finalize the v4.1 version of the WZDx specification for official release.

The `fixes/release-v4.1` branch will be updated as necessary fixes are found in reviewing #335. Once review of #335 is complete this PR can be completed and then #335 can be merged, marking the official v4.1 release.


j-d-b:
The example files FeedInfo version needs to be changed from 4.0 to 4.1

sknick-iastate:
> The example files FeedInfo version needs to be changed from 4.0 to 4.1

I have this updated in all of the examples pull request

j-d-b:
@sknick-iastate understood, thank you, but if our plan is to release v4.1 before merging the examples PR, we need to update it here as well or else the examples will be incorrect (not valid 4.0) until the examples PR is merged.

j-d-b:
I'm leaning toward preferring removing all the TDx references because they're not correct if the object/enum is duplicated in the TDx spec. The references note that the type is "used by" the TDx specification, but since the TDx spec has a copy of the type, the one in this repo/in WZDx isn't used by TDx, they just happen to be the same.

j-d-b:
The schema `$id`s and `$ref`s need to be updated to use the main branch in advance of merging into main

j-d-b:
> The example files FeedInfo version needs to be changed from 4.0 to 4.1

Completed

> The schema $ids and $refs need to be updated to use the main branch in advance of merging into main

Completed

j-d-b:
I reviewed every file with at least a skim except the examples as they will be reviewed and updated in #339.

sknick-iastate:
I went through almost all of the work zone, detour and device spec-content and looks good to me.

sknick-iastate (approved):

jlarsonOmahaNE (approved):

mark-mockett:
> I'm leaning toward preferring removing all the TDx references because they're not correct if the object/enum is duplicated in the TDx spec. The references note that the type is "used by" the TDx specification, but since the TDx spec has a copy of the type, the one in this repo/in WZDx isn't used by TDx, they just happen to be the same.
> 
> I think the references should only stay if the types are not duplicated in TDx and if TDx just includes links to WZDx rather than containing copies.

Removed all references to TDx from 'used by' sections

PR #344, opened 2022-10-05
Fix Work Zone Feed example GeoJSON files

j-d-b:
All example files for the WZDx Work Zone Feed in main are failing validation against the schema because some boolean properties are provided as strings in the example GeoJSON.

This PR updates all example files to change the strings to booleans—all example files then validate against the schema.


j-d-b:
See #343, which noted this issue.

mark-mockett (approved):

PR #349, opened 2022-10-07
Prep release v4.2 branch

mark-mockett:
Copy recent changes to main branch, add new schemas for v4.2, and scrub refs to v4.1 in documentation. 

Note: once new v4.1 examples are done, those may need to be brought in. Also, I didn't merge the changes to main in, so there may not be a clean commit history when we merge back in to main.


mark-mockett:
> I think we should update all the example json files to use "4.2" for the version on the FeedInfo. Otherwise looks good, thanks!

Done!

j-d-b (approved):

PR #358, opened 2022-10-31
Clarify use of LineString and MultiPoint in road event geometry

### LINKED ISSUE #194: Clarification on geometry requirements
### Taken from Discussion https://github.com/usdot-jpo-ode/wzdx/discussions/193

<div type='discussions-op-text'>

<sup>Originally posted by **Dunge** August  9, 2021</sup>
 ## Summary

The current geometry requirements are vague. It only states it needs to conform GeoJSON specifications, and is described as:

> The geometry of the road event. The Geometry object's type property MUST be LineString or MultiPoint. LineString allows specifying the entire road event path and should be preferred. MultiPoint should be used when only the start and end coordinates are known.

Questions:
* Can an event be composed of a single point? i.e. we know the start, but not the end. Can we push the same coordinate twice in MultiPoint mode?
* Does a LineString event needs to follow the road curvatures or can be an approximation?

 ## Motivation
I will expend on the second question.
When generating a LineString, we have a few options. Especially in the context of automating WZDx generation based on device GPS positions.

1. Draw on the map (the best case, require manual entry)
2. Automate and receive the path via mapping services like for example Google Maps based on the GPS waypoints.
3. Automate and generate lines going from one GPS position to the other without following the road curvature.

We used the **second method** for a while, but found it to be too unstable since it can lead to incorrect information in multiple situations:
* GPS is imprecise and appear in the other lane direction, or in a lane that goes near an highway exit.
* Mapping service detect a road closure and returns a path trying to avoid it by taking a detour.
* Mapping service is not aware of a temporary road placement and still use the normal road position.

This lead to obviously incorrect results like this
image

So this lead us to change to a simpler and more direct solution, the **third method**. Only draw lines from one validated GPS position to the other.

In this case as soon as the road have a bit of a curve, it will lead to lines "cutting" over places where there is no road:
image

I can see a similar representation in the TxDOT feed currently being published, where the workzone is on the I-635 but the LineString representation goes straight between two points:
image

Is this an acceptable output?
Of course, it depends on the feed consumers, and where they plan to display that geometry information. In some situations it could be accepted, but I believe on a public website this wouldn't look good, and for an automated vehicle it could be dangerous..

 ## Proposed changes

Unfortunately, we are better just to stop trying to publish a path that can't be confirmed to be correct. Since LineString geometry representation are kinda impossible to be attained automatically based solely on roadside devices GPS positions with a 100% certainty and following road curvatures, I believe MultiPoint would be a better fit in this case.

This lead to my proposed change. Remove the line from above description "MultiPoint should be used when only the start and end coordinates are known.". I would also remove the rule requiring MultiPoint to be only composed of pair and would allow a single RoadEvent to be composed of as many MultiPoint as possible, as long as they are representing the same workzone event.

The reason being that if you have a workzone with hundreds of sensor points, but the same information for the RoadEvent composing the whole workzone, it would be a bit irrational to split the workzone into hundreds of small events composed of start/stop pairs while the rest of the information remains the same and get duplicated leading to a huge output. It would be much better to have a single RoadEvent entry with the hundreds of points representing valid positions on the workzone, without any path between them.</div>

j-d-b:
This PR resolves #194 by updating the description and notes cells for the [RoadEventFeature](https://github.com/usdot-jpo-ode/wzdx/blob/main/spec-content/objects/RoadEventFeature.md).

Specifically, the following changes are proposed:

- Add the following sentences to the Description column for the RoadEventFeature `geometry`:

> The order of coordinates is meaningful: the first coordinate is the first (furthest upstream) point a road user encounters when traveling through the road event. If a data producer has three or more coordinates that are on the road event path, a `LineString` geometry should be used because it indicates that the points are ordered.

- Add the following sentence to the Notes column for the RoadEventFeature `geoemtry`:

> Feed producers should aim to provide a higher density of points for sections of roadways with curves so that data consumers can more easily match road events to their road network.


mark-mockett (approved):

spec-content/objects/RoadEventFeature.md
```diff
@@ -7,7 +7,7 @@ Name | Type | Description | Conformance | Notes
...
+`geometry` | GeoJSON [Geometry](https://tools.ietf.org/html/rfc7946#section-3.1) object with `type` of [LineString](https://tools.ietf.org/html/rfc7946#section-3.1.4) or [MultiPoint](https://tools.ietf.org/html/rfc7946#section-3.1.3) | The geometry of the road event. The Geometry object's `type` property MUST be [LineString](https://tools.ietf.org/html/rfc7946#section-3.1.4) or [MultiPoint](https://tools.ietf.org/html/rfc7946#section-3.1.3). `LineString` allows specifying the entire road event path and should be preferred. `MultiPoint` should be used when only the start and end coordinates are known. The order of coordinates is meaningful: the first coordinate is the first (furthest upstream) point a road user encounters when traveling through the road event. If a data producer has three or more coordinates that are on the road event path, a `LineString` geometry should be used because it indicates that the points are ordered. | Required | Feed producers should aim to provide a higher density of points for sections of roadways with curves so that data consumers can more easily match road events to their road network. This is a GeoJSON property.
```
jlarsonOmahaNE, line 10: does it make more sense to say "and is preferred" not "should be preferred"?

sknick-iastate (approved):

spec-content/objects/RoadEventFeature.md
j-d-b: Yes, but note that the "should be preferred" language is what is in the spec currently.

jlarsonOmahaNE: Cool, yeah I wasn't sure if you wanted to keep it the same. I think how you have it now is good too.

jlarsonOmahaNE (approved):

